PR #2, opened 2015-08-25
Ttree merge 2

kmcdermo:
Hi @dan131riley  @osschar  @cerati @pwittich @srlantz,

Here is the beast of a pull request I am submitting.  This can be automatically merged, but again, we can discuss changes here.

I can also make a README on how to use the validation tools I developed.  

For starters, ./main still spits out a valtree.root if ROOT is enabled for SMatrix code.  There are four trees:
tree_br: much like Giuseppe's old branching studies, with some upgrades
efftree: filled once per sim track, used for making efficiency plots, duplicate rates, pulls, resolutions, etc
faketree: filled once per seed, used for making fake rate plots (could also be used for pulls, resolutions)
configtree: branches for most of the stuff in Config.h 

once ./main is run, use:

root -l -b -q runValidation.C 

to produce a bunch of plots.  Takes about 20 seconds to produce them all.  runValidation.C allows you to specify the input file, output directory, and if you want to save the plots also as pdf/png/etc.  all plots stored in output root file with same name as output dir in the output dir.

Comments in TTreeValidation.cc explain the conventions for flags and filling.  Lots of ROOT branches to explore! (Check my new slides here as a demonstration of how convenient the branches are in making quick checks: https://www.dropbox.com/s/ec735zgy673i3j8/new_geo_losthits.pdf?dl=0)

Lastly, if you want to compare between "releases", use smatrixPlotComparison.C (specify ins/outs) to quickly compare efficiency/FR/DR/nHits between releases.  Could add pulls simply.  

  Peace,
Kevin


Config.h
```diff
@@ -0,0 +1,134 @@
...
+#define MAX_HITS 10
+#endif
+
+#ifdef USE_MATRIPLEX
+
+  #ifndef MPT_SIZE
+    #ifdef __MIC__
+      #define MPT_SIZE 16
+    #else
+      #define MPT_SIZE 1
```
cerati: let's keep the default of 8

Config.h
```diff
@@ -0,0 +1,134 @@
...
+    #endif
+  #endif
+
+  #ifndef NUM_THREADS
+  #define NUM_THREADS 1
+  #endif
+
+  #ifndef NUM_THREADS_SIM
+  #define NUM_THREADS_SIM 1
```
cerati: also here keep the default of 60

Hit.h
```diff
@@ -174,43 +160,29 @@ class Hit
...
-  const SMatrixSym33& error()      const {return state_.errors;}
+  const SVector3&  position()  const {return state_.parameters;}
+  const SMatrixSym33& error()  const {return state_.errors;}
+  const SVector3& parameters() const {return state_.parameters;}
+  const float r() const {return std::sqrt(getRad2(state_.parameters.At(0),state_.parameters.At(1)));}
+  const float x() const {return state_.parameters.At(0);}
+  const float y() const {return state_.parameters.At(1);}
+  const float z() const {return state_.parameters.At(2);}
+  const float phi() const {return getPhi(state_.parameters.At(0),state_.parameters.At(1));}
+  const float eta() const {return getEta(r(),z());}
```
cerati: I do not see the point for these floats to be const float

Hit.h
```diff
@@ -174,43 +160,29 @@ class Hit
...
 
   const MCHitInfo& mcHitInfo() const {return mcHitInfo_;}
-  int mcTrackID() const {return mcHitInfo_.mcTrackID_;}
-  int layer() const {return mcHitInfo_.layer_;}
-  int ithLayerHit() const {return mcHitInfo_.ithLayerHit_;}
-  int hitID() const {return mcHitInfo_.mcHitID_;}
+  unsigned int mcTrackID() const {return mcHitInfo_.mcTrackID_;}
+  unsigned int layer() const {return mcHitInfo_.layer_;}
+  unsigned int ithLayerHit() const {return mcHitInfo_.ithLayerHit_;}
+  unsigned int hitID() const {return mcHitInfo_.mcHitID_;}
```
cerati: @osschar is going to complain against unsigned ints :-)

Makefile.config
```diff
@@ -17,26 +17,26 @@
 
 # 1. Use ROOT or not (never used on MIC)
 # Comment out to disable root ("yes" is not relevant)
-#WITH_ROOT := yes
+WITH_ROOT := yes
```
cerati: let's keep the old default

Simulation.cc
```diff
@@ -101,10 +118,10 @@ void setupTrackByToyMC(SVector3& pos, SVector3& mom, SMatrixSym66& covtrk, HitVe
     // can I do what I did, generate one (theta_space) and rotate it by another random numbers
     const float z1 = g_gaus(g_gen);
     const float z2 = g_gaus(g_gen);
-    const float phismear = g_unif(g_gen)*TMath::TwoPi(); // random rotation of scattering plane
+    const float phismear = g_unif(g_gen)*Config::TwoPI; // random rotation of scattering plane
     const float X0 = 9.370; // cm, from http://pdg.lbl.gov/2014/AtomicNuclearProperties/HTML/silicon_Si.html
     //const float X0 = 0.5612; // cm, for Pb
-    float x = 0.1; // cm  -assumes radial impact. This is bigger than what we have in main
+    float x = 0.0025; //.1 cm  -assumes radial impact. This is bigger than what we have in main
```
cerati: what is this parameter? why has it changed? should it go to Config.h?

Track.h
```diff
@@ -56,32 +57,44 @@ class Track
...
+  const float momPhi() const {return getPhi(state_.parameters[3],state_.parameters[4]);}
+  const float momEta() const {return getEta(pt(),pz());}
+
+  // track state momentum errors
+  const float ept()     const {return std::sqrt(std::abs(getRadErr2(state_.parameters[3],state_.parameters[4],state_.errors[3][3],state_.errors[4][4],state_.errors[3][4])));}
+  const float epx()     const {return std::sqrt(state_.errors[3][3]);}
+  const float epy()     const {return std::sqrt(state_.errors[4][4]);}
+  const float epz()     const {return std::sqrt(state_.errors[5][5]);}
+  const float emomPhi() const {return std::sqrt(std::abs(getPhiErr2(state_.parameters[3],state_.parameters[4],state_.errors[3][3],state_.errors[4][4],state_.errors[3][4])));}
+  const float emomEta() const {return std::sqrt(std::abs(getEtaErr2(state_.parameters[3],state_.parameters[4],state_.parameters[5],state_.errors[3][3],state_.errors[4][4],state_.errors[5][5],state_.errors[3][4],state_.errors[3][5],state_.errors[4][5])));}
```
cerati: I am not enthusiast about this "const float" campaign...

Track.h
```diff
@@ -109,12 +127,20 @@ class Track
...
   HitVec hits_;
-  HitVec initHits_;
-  HitIdxVec hitIdxVec_;
   float chi2_;
+  unsigned int mcTrackID_;
+  unsigned int nHitsMatched_;
+  unsigned int seedID_;
+  unsigned int duplicateID_;
+  bool isDuplicate_;
+  HitIdxVec hitIdxVec_;
```
cerati: you are actually adding private members to track, this is not good as it will increase the size. 
we should merge asap also the data format reduction from my branches

mkFit/MkFitter.cc
```diff
@@ -501,19 +501,23 @@ void MkFitter::FindCandidates(std::vector<Hit>& lay_hits, int firstHit, int last
...
 
 void MkFitter::GetHitRange(std::vector<std::vector<BinInfo> >& segmentMapLay_, int beg, int end,
-                           const float etaDet, int& firstHit, int& lastHit)
+                           int& firstHit, int& lastHit)
 {
 
     int itrack = 0;
 
 
+    // what exactly is this function doing?  why are you calling the first simtrack in the event to get the eta??
```
cerati: not sure if this kind of comments should go into the devel branch.
we should rather discuss in person what these methods do

mkFit/MkFitter.cc
```diff
@@ -631,9 +635,9 @@ void MkFitter::SelectHitRanges(BunchOfHits &bunch_of_hits)
 #endif
 
     phiBinMinus = std::max(0,phiBinMinus);
-    phiBinMinus = std::min(Config::nPhiPart-1,phiBinMinus);
+    phiBinMinus = std::min(int(Config::nPhiPart-1),phiBinMinus);
```
cerati: can't we just make Config::nPhiPart an simple int ?

mkFit/buildtestMPlex.cc
```diff
@@ -12,7 +14,7 @@
 #include "ittnotify.h"
 #endif
 
-// #define PRINTOUTS_FOR_PLOTS
+#define PRINTOUTS_FOR_PLOTS
```
cerati: please restore the old default

Hit.h
dan131riley: @osschar, @cerati what's the complaint about unsigned ints?

Track.h
dan131riley: @cerati Agreed, making return-by-value scalars const doesn't serve any purpose

Track.h
dan131riley: @cerati @kmcdermo These should probably go into some kind of "track extra" structure, or in a specialized MCTrack class that refers to the Track class.  We should also discuss merging the data format reduction in the (currently mis-named) dsr -mpx-build branch, which has a stripped down Hit and Track class compatible with yours, but also an MCHit class that preserves the extra bookkeeping Kevin's validation wants.

Hit.h
osschar: I'm one of the guys who think, and feel rather strongly about it :) , that unsigned should only be used for bit-fields, register access, and for cases when the extra bit really really matters (e.g., as an object-id in a system with fast object turnaround (and where it is then used as hash/db key, not as an index)).

Makefile
```diff
@@ -39,6 +39,10 @@ distclean: clean
 	-rm -f ${AUTO_TGTS}
 	cd mkFit && ${MAKE} distclean
 
+squeakyclean: distclean
+	-rm -f *.optrpt
+	cd mkFit && ${MAKE} squeakyclean
+
```
osschar: Just add this rm to distclean.

Makefile.config
osschar: Yes, I agree ... keep old defaults, just add what needs to be added.

BinInfoUtils.h
```diff
@@ -0,0 +1,55 @@
...
+}
+	
+inline float upPhi(float phi){
+  while (phi <= -Config::PI) {phi+=Config::TwoPI;}
+  return phi;
+}
+
+inline float normalizedPhi(float phi) {
+  //  return std::fmod(phi, (float) Config::PI); // return phi +pi out of phase for |phi| beyond boundary! 
+  if (fabs(phi)>=Config::PI) {phi = (phi>0 ? downPhi(phi) : upPhi(phi));}
```
osschar: I'd go for std::abs(), everywhere.

Config.h
```diff
@@ -0,0 +1,134 @@
+#ifndef _config_
+#define _config_
+
+#include <algorithm>
+
+namespace Config{  
+  // math general --> from namespace TMath
+  static constexpr const float    PI    = 3.14159265358979323846;
```
osschar: static (internal linkage) is implied for const objects in a namespace -- so drop static.

constexpr implies const so you can use constexpr as the only specifier.

kmcdermo:
pushed some minor changes: restored defaults and put in Config::nLayers in more places.

Simulation.cc
kmcdermo: @cerati I believe it is related to the scattering somehow.  I must have changed it when I was messing with the beamspot studies.  Seeing as Mplex currently does not use scattering, I just changed it back to 0.1.  

In reality, float x = std::sqrt(getRad2(Config::beamspotX, Config::beamspotY)) -- I think, need to confirm with @pwittich 

And yes, I think x0 should also be moved to Config.h.  I will make a commit for that.

Config.h
kmcdermo: done

Config.h
kmcdermo: done

Hit.h
kmcdermo: @osschar @cerati  I will change these (and other unsigned ints I inserted) to ints.  This will be the biggest change.  I will get back to you when it is done.

Makefile.config
kmcdermo: done

Makefile
kmcdermo: done

mkFit/buildtestMPlex.cc
kmcdermo: done

mkFit/MkFitter.cc
kmcdermo: yes, will do this once I change all unsigned ints to regular ints

mkFit/MkFitter.cc
kmcdermo: done... will commit shortly

BinInfoUtils.h
kmcdermo: @osschar  I can replace fabs with std::abs() in this file, but for example: ./mkFit/MkFitter.cc:441 or ./mkFit/PropagationMPlex.cc:600 has fabs().  Do we want std::abs() everywhere, including Math/ ?

Additionally, there are sqrt, std::sqrt, std::sqrtf .... what to do with those?

Hit.h
kmcdermo: @osschar @cerati @dan131riley There are some unsigned ints that are in various places not from my changes.

Examples: 
https://github.com/cerati/mictest/blob/devel/ConformalUtils.cc#L21
https://github.com/cerati/mictest/blob/devel/Propagation.cc#L255-L256
https://github.com/cerati/mictest/blob/devel/Geometry.h#L20-L22
Math/ (all over the place)

Should I change these as well?  I feel like ConformalUtils, Propagation, Geometry makes sense, changing Math/ sounds like a bad idea.

dan131riley: I wouldn't go doing a mass conversion since that could introduce bugs (I also generally don't agree with Matevz about unsigned vs. signed (unsigned is simple and forms a group :), but not particularly strongly.

-dan

cerati: Hi Kevin, 
I do not think you should replace all "unsigned " with "" in all parts of the code, but rather undo the additions of "unsigned " that you made. 
Giuseppe

Config.h
kmcdermo: done

Track.h
kmcdermo: removed first const where appropriate (i.e. for functions I previously defined const type const {return state.value} --> only for by value... kept const type const for return by reference

Hit.h
kmcdermo: Changed the consts appropriately in Hit.h, Track.h, and Event.h, see below.

Simulation.cc
kmcdermo: Moved x and x0 to config.  since "x" is everywhere in our code, I made the parameter x previously in Simulation.cc as Config::xr (the r for the radial impact from the comment by Peter)

kmcdermo:
Commit messages match the minor commit diffs.

BinInfoUtils.h
kmcdermo: I changed only the fabs() in BinInfoUtils.h to std::abs().  As we discussed in the meeting, we will need to make a commit fully changing all fabs() and abs() to std::abs(), as well as sqrt, sqrtf, etc.

Track.h
kmcdermo: @cerati @dan131riley @osschar 

I am a bit stuck here for a number of reasons.  I am now remembering why I decided to explicitly make some ints change to unsigned ints.
If you look above, you will see I changed layer() to unsigned, to match what Geometry.h/.cc yields for the layer info.  

However, the eta binning structures between SMatrix and Matriplex use eta bins as either ints or unsigned.  Since we would rather port the overlap to SMatrix, it would make sense to move the bins in Smatrix to ints and not unsigned.

I have a working version of the smatrix code changed appropriately for ints.  THis meant changing the Geometry code to ints, some things in Propagation, etc.  Math/ was left unchanged.  Additionally, anything that does not have an explicit comparison with my changes was left as is (some unsigned ints remaining).  

I checked the validation of the smatrix, and the results of changes to many ints to my version is identical.  

under: ttree-merge-2-int

cerati: Hi @kmcdermo, are you suggesting to merge this one as it is and then go through ttree-merge-2-int?

kmcdermo: Hi @cerati , yes, this branch is all ready to go, except for unsigned int -> int.  To reiterate, I still need some guidance on what to do with them in Geometry / binning.

PR #5, opened 2015-09-23
Refactor track building 

osschar:
Refactor track building so that best-hit and combinatorial version reuse as much code as possible.

For now this is implmented as a class declared and defined inside of
buildtestMPlex.cc. We should eventually take this out but I'd prefer to see
what kind of trouble we get into when doing the feature merge from the
exploration branches.


cerati:
looks good by eye. did you already check if it compiles and run properly?

PR #7, opened 2015-09-28
More Validation Tools, Cleanups of seedtest with Configs, discard normalizeEta, change getPhiPartition

kmcdermo:
Hi All,

I have a number of changes I would like committed that do not affect physics performance (although may affect timing performance of MkFit if my BinInfoUtils.cc will be used in mkFit). 

1) I have included more validation plots (hurray!). 

2) I cleaned up seedtest.cc as Giuseppe said he might start playing around with it.  Mainly, this just meant replacing hardcoded consts with consts in Config.h

3) I did away with normalizedEta() in BinInfoUtils (and where it is called in the SMatrix code).  The reason is something sinister... floating point comparisons that fail (when 1 != 1).  I kept getting really annoying crashes when this happened.  Instead, I have chosen to just bin directly any output of the bin calculation that is outside the bounds of our virtual detector eta bins.  

4) Need the same bin check in getPhiPartition as well... had a nasty corner case of pi != pi that caused code to crash as it returned a bin number = Config::nPhiPart (which is +1 outside the range of 0 - Config::nPhiPart).  Would be nice if we could use module on phi wrapping, but I have checked this before, and anything more than a few radians away from phi < |pi| it gets progressively worse with increase in distance from pi.

N.B. This PR incorporates all the PRs from before (i.e. can be automatically merged).

### DIFF
--- a/BinInfoUtils.h
+++ b/BinInfoUtils.h
@@ -32,13 +32,16 @@ inline unsigned int getPhiPartition(float phi){
   //assume phi is between -PI and PI
   //  if (!(fabs(phi)<Config::PI)) std::cout << "anomalous phi=" << phi << std::endl;
   //  const float phiPlusPi  = std::fmod(phi+Config::PI,Config::TwoPI); // normaliztion done here
-  const float phiPlusPi  = phi+Config::PI; 
-  const unsigned int bin = phiPlusPi*Config::fPhiFactor; // would have to change this to just a regular int (no const) if want to use commented lines below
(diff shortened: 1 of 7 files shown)


kmcdermo:
@dan131riley @cerati @osschar @pwittich 

This PR can now safely be merged.  None of this code depended on the slimmed data structures (almost exclusively validation + some minor cleanups in the SMatrix version).  I would suggest merging this before we move to building the "Track Extra" class.

PR #8, opened 2015-10-06
Merge slimmed Hit and Track into devel

dan131riley:
This certainly has bugs, mkFit efficiency is low and main seems slow.  It also disables validation and seeding code that still needs conversion.  However, it at least gets the bulky changes out for review.

### DIFF
--- a/Event.cc
+++ b/Event.cc
@@ -12,7 +12,7 @@
 #include "tbb/tbb.h"
 #endif
 
-static bool sortByPhi(const Hit& hit1, const Hit& hit2)
+inline bool sortByPhi(const Hit& hit1, const Hit& hit2)
 {
   return hit1.phi()<hit2.phi();
 }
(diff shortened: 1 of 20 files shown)


cerati:
Hi @dan131riley, from a visual inspection the changes look fine to me. Let me know if you need help debugging the low efficiency issue.

dan131riley:
I found two large problems, one with mapping the hits on the simtrks correctly, which affected sMatrix because we sort the hits into layerHits; added a new function to remap from MC truth hits to the simulated hits.  Second problem was a relatively trivial issue with offsets into the Hit structure getting messed up.

I also ported over some "end - itrack" fixes from the mt2 branch.  I believe MkFitter::FindCandidates() still has problems in this area, I added a couple of assertions (currently commented out) that trigger when the # of tracks isn't a nice number, where it tries to add hits to non-existent seedidx entries.

Efficiency of MkBuilder::FindTracks() is still lower than expected, but not as dramatically, so I think this should be merged while work on that and validation continues.

PR #19, opened 2016-01-08
Change HitID types from short to int

dan131riley:
Source of 30k event limit.  I had thought this would be used more than it is, so I over-optimized.  Only used in MC generation, so doesn't matter so much.


cerati:
Thanks!

PR #31, opened 2016-02-06
Reduce clone thread to really only do cloning

osschar:


cerati:
benchmarking started... will merge after checking the result tomorrow

osschar:
OK, I did what we discussed. Standard and clone-engine versions now produce identical results down to +- 2 tracks.

PR #33, opened 2016-03-03
Revive save track

cerati:
- add back functionality for saving and reading back in memory sim tracks from a file
- this also speeds up significantly the benchmarking

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -2,6 +2,9 @@
 
 namespace Config
 {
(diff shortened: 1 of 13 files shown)


cerati:
benchmarking plots at http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-03-03-PR33/

slightly faster, most likely due to the reduction in size of the containers

merging...

PR #36, opened 2016-03-08
Super debug mode: SMatrix 

kmcdermo:
This commit is only for SMatrix.

This version allows a test of building in SMatrix to follow the propagation and update track states of single track events with no skipped layers.

This tool is enabled with a simple ifdef, which sets a bool in Config.h. This forces the code to compile everything, then only perform the other validation when true.

Closed previous PR by accident.  Sorry for the duplication.

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -2,8 +2,14 @@
 
 namespace Config
 {
+
+#ifdef SUPERDEBUG // to avoid include of Config.h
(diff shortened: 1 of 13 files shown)


cerati:
No need to check benchmarking since mkFit is not affected.
Kevin checked SMatrix version compiles, runs and produces consistent results.

PR #37, opened 2016-03-10
Ports fittest to CUDA

mpbl:
I have tried to squash all commits in as few super-commits as possible.


cerati:
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-03-18-PR37/

PR #38, opened 2016-03-11
Add commandline parametrs for nEvents and nTracks.

osschar:


cerati:
I guess this one is simple enough to skip the benchmarking.
But I think we should change the benchmarking scripts to use this new feature.

PR #39, opened 2016-04-08
Kalman filter calculations rework

cerati:
- rework of kalman calculations (both in Matriplex and SMatrix), to use hit coordinates on tangent plane and "polar" coordinates for kalman update
- add curvilinear jacobian (as cmssw), not used and not optimized
- go back to iterative jacobian by default
- add option to read seed collection from cmssw input "readCmsswSeeds"
- protections against non finite eta values
- fix small issue with wrapping in SelectHitRanges causing small inconsistency in output for different MPT_SIZE
- update of Math/MatrixRepresentationsStatic.h from recent ROOT version, looks like SMatrix has now reduced the object size as well
- other minor fixes

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -18,4 +18,5 @@ namespace Config
   int   finderReportBestOutOfN = 1;
 
   bool  useCMSGeom = false;
+  bool  readCmsswSeeds = false;
 }
--- a/Config.h
+++ b/Config.h
@@ -79,8 +79,8 @@ namespace Config
   constexpr int Niter = 5;
   constexpr float Bfield = 3.8112;
   constexpr bool doIterative = true;
-  constexpr bool useSimpleJac = true;  
-  constexpr bool useCurvJac = false;
+  constexpr bool useSimpleJac = false;
+  constexpr bool useCurvJac   = false;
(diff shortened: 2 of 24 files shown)


kmcdermo:
Hi @cerati , why are there changes to the Math/  directory?  Just glancing at the changes, I have a feeling this is because of multiplying symmetric matrices by non-symmetric matrices (the compiler getting all mad).  I had this problem quite a bit when investigating the KF update with symmetric vs. nonsym matrices... If you found a work-around, +1 for sure.

dan131riley:
Looks like it's just updated from a newer version of ROOT (which is good, what's in Math should stay compatible with ROOT since that's where it came from, and we don't want to lose the ability to compile with the ROOT versions of those headers).

-dan

cerati:
Hi Kevin, do you mean the changes in  Math/MatrixRepresentationsStatic.h? It is new copy from ROOT6, which actually reduces the size of SMatrix to the same as the plain array of floats - basically looks like they also realised they had unecessary data members

cerati:
Benchmarking results: http://uaf-7.t2.ucsd.edu/~cerati/mictest/2016-04-08-PR39/

PR #40, opened 2016-04-11
TBB track builder

dan131riley:
Adds track builder threading with TBB.  Factors out the core of track building into separate routing shared by the OMP and TBB builders.  Adds new flags for the TBB builder and number of seeds per task with TBB.  Adds TBB support to the benchmark suite.  Moves seed fit and initial binning out of the time section, so all builders get somewhat faster.


cerati:
Hi @dan131riley, are you going to run the benchmarking script or you prefer me to do it?

dan131riley:
@cerati, put the initial results here:

http://www.lepp.cornell.edu/~dsr/mic-track/2016-04-11-PR40/

re-running now with a couple of small changes, shouldn't be any different.

-dan

cerati:
@dan131riley, thanks.
Once this test is done, can you point me to the location on phiphi so that I can copy the pngs to my usual area (I'd keep all the plots in the same place)?
Also, I thought we wanted to get rid of OMP now that we established that TBB works better - is this planned in a followup pull request?

srlantz:
@cerati, I'm not sure we want to get rid of OpenMP entirely. The fact is, so far we have been using OMP in a relatively simple way--I think just in order to get started. I believe OMP 3.0 has task-based features that may not be too different from what we're now doing with TBB. (OMP 4.0 seems more focused on accelerators and SIMD, though it does introduce some task enhancements, such as task dependency and thread affinity... I must admit I have not kept up with all the latest developments... the OpenMP 4.0 document is 250+ pp. long, while 1.0 was just 40 pp.)

Anyway, given that OMP is arguably more portable, we may not want to let go of it just yet. Or I suppose we could ditch OMP for now, then start over with it at a later time, assuming we learn a lot from the TBB experience. Probably we all agree that this is a low-priority area of development, given the more pressing performance issues we'd like to address in the near term.

One good reason I can think of for tossing OMP is if it constrains our code structure in some bad way, so that it inhibits future development.

dan131riley:
@cerati,

see /home/dsr/mictest-mt on phiphi.  wrt removing OMP, I think it's useful to have a version with both OMP and TBB so we have a direct comparison with a continuous record of performance changes.  We also still need TBB versions of (at least) best hit and the fitter, so I think the development plan is to do those next, and then after doing a PR for that we can look at removing the OMP versions.  I am also doing some refactoring to minimize the code duplication, so the chunks of code that are OMP-specific are getting smaller.

@srlantz,

I'd argue that TBB is more portable!  TBB works with any C++11 (or later) compiler, all it needs is features that are in the C++ standard.  OpenMP needs special compiler support, so (for example) TBB works with the standard system compiler on the Mac, but for OpenMP I need to install pre-release versions of LLVM.  We've also got a mandate from CMS that anything we expect to incorporate into the standard CMS processing change has to use TBB, as that's what the CMSSW framework uses.

Also, even the latest versions of OpenMP is based on C++98, so its interaction with C++11 features isn't defined.  I find this very constraining, as I find C++11/C++14 much more expressive than C++98.

-dan

srlantz:
@dan131riley, good counter-arguments, thank you for making these points.

cerati:
for the record: copied to http://uaf-7.t2.ucsd.edu/~cerati/mictest/2016-04-13-PR40/

dan131riley:
Hi @cerati,

sorry, could you recopy those from /home/dsr/mictest-mt/2016-04-13-PR40?  I was doing some further testing and forgot you still needed to copy those, so if you look at the host_th_speedup.png and host_th_time.png you copied, you'll see those are actually from a later version with the best hit converted to TBB (and showing an improvement in scaling).  I've put the correct versions back to /home/dsr/mictest-mt/2016-04-13-PR40 so that can't happen again.

-dan

cerati:
done, thanks
g.

dan131riley:
Hi @cerati,

I added a couple of changes to PR#40 to fix problems with setting the number of threads.  This bug only hit when -read was not used, so it doesn't affect the test results.  Are you going to go ahead with the merge?

-dan

cerati:
I thought I already merged, but I guess I did not push the button, doing it now!

PR #43, opened 2016-04-29
Debug cleanup

dan131riley:
Cleanup and standardize debug statements.  Touches lots of files but no algorithm changes, all changes are inside DEBUG defines.  Uses dprint()/dcall()/dprintf() macros to try to eliminate "#ifdef DEBUG" where reasonable, standardizes on "debug" bool (changing "dump" to "debug" in places), and adds a C++11 mutex for all debug statements (replacing the OMP mutex in MkBuilder).  Most extensive changes are in MkBuilder due to consolidation of cut&paste print statements into print routines, changes to other files are all pretty straightforward.  Has been test compiled with (and without) -DDEBUG with icc and clang.  Since there are no algorithmic changes, did not run benchmarks.

### DIFF
--- a/seedtest.cc
+++ b/seedtest.cc
@@ -187,7 +187,7 @@ void buildSeedsByRoadTriplets(TrackVec& evt_seed_tracks, TrackExtraVec& evt_seed
   if (debug){
     dprint("Hit Pairs");
     for(auto&& hitPair : hitPairs){
-      printf("ilay0: %1u ilay1: %1u  \n",
+      dprintf("ilay0: %1u ilay1: %1u  \n",
  	     ev.simHitsInfo_[evt_lay_hits[0][hitPair[0]].mcHitID()].mcTrackID(),
  	     ev.simHitsInfo_[evt_lay_hits[1][hitPair[1]].mcHitID()].mcTrackID()
 	     );
@@ -216,7 +216,7 @@ void buildSeedsByRoadTriplets(TrackVec& evt_seed_tracks, TrackExtraVec& evt_seed
   if (debug){
     dprint("Hit Triplets");
     for(auto&& hitTriplet : hitTriplets){
-      printf("ilay0: %1u ilay1: %1u ilay2: %1u \n",
+      dprintf("ilay0: %1u ilay1: %1u ilay2: %1u \n",
  	     ev.simHitsInfo_[evt_lay_hits[0][hitTriplet[0]].mcHitID()].mcTrackID(),
  	     ev.simHitsInfo_[evt_lay_hits[1][hitTriplet[1]].mcHitID()].mcTrackID(),
  	     ev.simHitsInfo_[evt_lay_hits[2][hitTriplet[2]].mcHitID()].mcTrackID()
--- a/seedtest.cc
+++ b/seedtest.cc
@@ -217,7 +217,7 @@ void buildSeedsByRoadTriplets(TrackVec& evt_seed_tracks, TrackExtraVec& evt_seed
     for(auto&& hitTriplet : hitTriplets){
       printf("ilay0: %1u ilay1: %1u ilay2: %1u \n",
  	     ev.simHitsInfo_[evt_lay_hits[0][hitTriplet[0]].mcHitID()].mcTrackID(),
- 	     ev.simHitsInfo_[evt_lay_hits[1][hitTriplet[1]].mcHitID()].mcTrackID()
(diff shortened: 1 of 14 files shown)


seedtest.cc
```diff
@@ -324,7 +324,7 @@ void buildHitTripletsCurve(const std::vector<HitVec>& evt_lay_hits, const BinInf
     const auto phiBinMinus = getPhiPartition(negPhi);
     const auto phiBinPlus  = getPhiPartition(posPhi);
 
-#ifdef DEBUG
+#ifdef BROKEN_DEBUG
     const float lay2phi = evt_lay_hits[2][ev.simTracks_[ev.simHitsInfo_[hit0.mcHitID()].mcTrackID()].getHitIdx(2)].phi();
     dprint("lay0 phi: " << hit0.phi() << " lay1 phi: " << hit1.phi() << std::endl <<
 	   "negPhi: " << negPhi << " lay2 phi: " << lay2phi << " posPhi: " << posPhi << std::endl <<
```
kmcdermo: @dan131riley What about this particular section is broken?  I presume the BROKEN_DEBUG implies something is non-standard with how you have reorganized debugging.  I can fix it -- let me know!

kmcdermo:
@dan131riley How many places still use #ifdef DEBUG, and why would it not be reasonable to remove the ifdef?

seedtest.cc
dan131riley: It uses ev which isn't in scope, probably hasn't been compiled since some
reorganization that moved it to where ev isn't available:

seedtest.cc:328:43: error: use of undeclared identifier 'ev'
    const float lay2phi = evt_lay_hits[2][ev.simTracks_[ev.simHitsInfo_[...
                                          ^
seedtest.cc:328:57: error: use of undeclared identifier 'ev'
    const float lay2phi = evt_lay_hits[2][ev.simTracks_[ev.simHitsInfo_[...

dan131riley:
On Apr 30, 2016, at 07:54, Kevin McDermott <notifications@github.com> wrote:
> 
> @dan131riley How many places still use #ifdef DEBUG, and why would it not be reasonable to remove the ifdef?

@kmcdermo,

There's a bunch of places that have some kind of control structure
like an if or loop.  Where we have that kind of control structure,
it isn't enough for the print statements to disappear when DEBUG
isn't defined, we want the whole control structure to disappear (or
be completely certain that the compiler will notice it doesn't do
anything and toss it away).  Where we have cut&paste print statements
(e.g., in MkBuilder and fitttest) I've moved those into functions,
and elide the calls to those functions via dcall(), but it didn't seem
worth it for the things that aren't repeated.  I suppose there's other
ways those control structures could be made to disappear (maybe via
templates and lambdas), but again it didn't really seem worth the
effort just to get rid of the ifdef.

-dan

cerati:
Hi @dan131riley,
sorry for the late feedback!
I have not looked in detail at the code, but let me make sure one feature I find useful is still possible.
While debugging specific parts of the code, I ofter define DEBUG only in specific files. 
Is the new version still compatible with this idea?
By the way, even if the changes are not algorithmic, I would still prefer the benchmarking is run on PR that are more than 1 line change...
Thanks,
Giuseppe

dan131riley:
@cerati,

Yes, you can turn on DEBUG only in specific files, the new change is that it has to defined before including Debug.h,

#define DEBUG // must be before include
#include <Debug.h>

However, that does turn the dprint/dcall/dprintf macros on for the entire file.  If you want printouts only for specific routines, you can set the global "bool debug" to false and create a local "bool debug = true;" for the scopes you want printouts enabled.

Supporting benchmarks are now in phiphi:/home/dsr/mictest-mt/2016-05-04-PR43

-dan

srlantz:
@dan131riley,

Maybe it would be good to edit Debug.h to include a comment or two about how to enable DEBUG in specific files and specific routines? Your instructions in this thread are simple enough and can be deduced, but for reference, it would be handy to have a quick summary in the code itself.

cerati:
@dan131riley thanks for the explanation and for the plots. 
I think the PR is good to go, let me know if you plan to add the instructions as requested by @srlantz before merging.

dan131riley:
@cerati,

I've added comments, thanks for the suggesting @srlantz

-dan

cerati:
for the record, benchmarking plots are at 
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-04-PR43/

PR #44, opened 2016-05-01
Conformal Fit Matriplex Port

kmcdermo:
Full port of conformal fit to Matriplex.  This is needed as a first step for seeding, as well as the initial input to the final fit over built tracks.  As seeding relies on eta/phi bin structures which are currently under development, this will be ported after that work is completed.  There are a number of places that this code can be optimized, most notably the redundancy of vector B (which is just a copy of the vector v), as well as replacing MPlexHV x, y, and z with just a single MPlexHH.  This redundancy appears as an artifact of the SMatrix version.  

I ran the benchmarking with all the edits, but turning off the CF:  https://kmcdermo.web.cern.ch/kmcdermo/before/
and after turning on the CF: https://kmcdermo.web.cern.ch/kmcdermo/after/
As expected, the nHits gets worse (already seen in SMatrix) when using the CF vs MC.  I am currently investigating to see if there is a smoking gun for the loss. 

This PR also contains the fixes to the SMatrix CF for the u vs v mapping from the phi structure discussed previously, as well as more branches for the super debug mode tree in SMatrix validation.

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -26,6 +26,8 @@ namespace Config
   bool  useCMSGeom = false;
   bool  readCmsswSeeds = false;
 
-  bool  super_debug = false;
   bool  cf_seeding  = false;
+  bool  cf_fitting  = false;
+
+  bool  super_debug = false;
 }
--- a/Config.h
+++ b/Config.h
@@ -17,6 +17,7 @@ namespace Config
   constexpr float TwoPI    = 6.28318530717958647692;
   constexpr float PIOver2  = Config::PI / 2.0;
   constexpr float PIOver4  = Config::PI / 4.0;
+  constexpr float PI3Over4 = 3.0 * Config::PI / 4.0;
   constexpr float InvPI    = 1.0 / Config::PI;
   constexpr float RadToDeg = 180.0 / Config::PI;
   constexpr float DegToRad = Config::PI / 180.0;
@@ -124,9 +125,9 @@ namespace Config
   constexpr float phierr049   = 0.0017; // 0.0017;
   constexpr float thetaerr049 = 0.0033; // 0.0031; 
   // parameters for layers 0,1,2 // --> ENDTOEND with "real seeding", fit is outward by definition, with poly geo
-  constexpr float ptinverr012 = 0.1234; // 0.1789;  -->old values from only MC seeds
-  constexpr float phierr012   = 0.0071; // 0170; 
-  constexpr float thetaerr012 = 0.0130; // 0.0137; 
+  constexpr float ptinverr012 = 0.12007; // 0.1789;  -->old values from only MC seeds
+  constexpr float phierr012   = 0.00646; // 0.0071 
+  constexpr float thetaerr012 = 0.01366; // 0.0130; 
 
   // matrix config
   // XXXX MT this should be renamed, made constexpr
--- a/Config.h
+++ b/Config.h
@@ -10,7 +10,6 @@ namespace Config
 {
   // super debug mode in SMatrix
   extern bool super_debug;
-  extern bool cf_seeding;
 
   // math general --> from namespace TMath
   constexpr float    PI    = 3.14159265358979323846;
@@ -85,7 +84,8 @@ namespace Config
   constexpr float seed_z0cut   = beamspotZ * 3.0; // 3cm
   constexpr float lay2Zcut     = hitposerrZ * 3.6; // 3.6 mm --> to match efficiency from chi2cut
   constexpr float seed_d0cut   = 0.5; // 5mm
-  
+  extern bool cf_seeding;
+
   // Config for propagation
   constexpr int Niter = 5;
   constexpr float Bfield = 3.8112;
@@ -128,6 +128,9 @@ namespace Config
   constexpr float ptinverr012 = 0.12007; // 0.1789;  -->old values from only MC seeds
   constexpr float phierr012   = 0.00646; // 0.0071 
   constexpr float thetaerr012 = 0.01366; // 0.0130; 
+
+  // config on fitting
+  extern bool cf_fitting;
 
   // matrix config
   // XXXX MT this should be renamed, made constexpr
--- a/ConformalUtils.cc
+++ b/ConformalUtils.cc
@@ -18,10 +18,21 @@ void conformalFit(const Hit& hit0, const Hit& hit1, const Hit& hit2, int charge,
   z[1]=hit1.position()[2];
   z[2]=hit2.position()[2];
 
-  float u[3],v[3];
-  for (unsigned int h=0;h<3;++h) {
-    u[h]=x[h]/(x[h]*x[h]+y[h]*y[h]);
-    v[h]=y[h]/(x[h]*x[h]+y[h]*y[h]);
+  float u[3],v[3]; // conformal points
+  float initphi = abs(atan2(y[1],x[1])); // use this to decide when to use x -> u or x -> v
+  bool xtou = (initphi<Config::PIOver4 || initPhi>Config::PI3Over4);
+
+  if (xtou){ // x -> u 
+    for (unsigned int h=0;h<3;++h) {
+      u[h]=x[h]/getRad2(x[h],y[h]);
+      v[h]=y[h]/getRad2(x[h],y[h]);
+    }
+  }
+  else { // x -> v
+    for (unsigned int h=0;h<3;++h) {
+      v[h]=x[h]/getRad2(x[h],y[h]);
+      u[h]=y[h]/getRad2(x[h],y[h]);
+    }
   }
 
   //R^2=a^2+b^2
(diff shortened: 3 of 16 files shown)


cerati:
Hi @kmcdermo,
two quick things:
- the version with the CF seeding is, if anything, slightly faster (?)... are you including the CF in the time benchmark?
- in general, we try to reduce as much as possible the number of temporaries in the mkFit code so that we avoid filling up the cache (this may come as a later cleanup of course or may not be relevant if you already include the CF in the time measurement)
Thanks,
Giuseppe

kmcdermo:
Hi @cerati ,

To your first question... I do not think so (at least for the building).  My understanding is that the building excludes the timing over the fit to seeds.  Is that correct @dan131riley , @srlantz , @osschar ?  Keep in mind that the CF seeding recovers less hits, so in building, this is why it is most likely just a bit faster (less candidates to process), if in fact the timing over the fit to seeds is excluded.  

As for the fitting test, the times go up and down for both before/after, I would say that is negligible (despite the fact that the CF is in fact included in the timing).

As you can see from the code, there are quite a few temporaries, and I even have a comment on it :).  I could of course just put explicitly everywhere in the calculation the input parameters, but I figured the temporaries simplify things... let me know what you think.

osschar:
Yes, we exclude fit_seeds from timing of building ... we're timing the track finding and seeds are supposed to come from "outside". If we actually need to do Kalman fit of the seeds that's another thing then.

How can you recover fewer hits? Shouldn't proper (well, not mc-based) seeding give us more initial seeds?

kmcdermo:
Hi Matevž,

At the moment, this is not the full real seeding, i.e. triplet finding +
CF.  The triplet finding we agreed we would port after the big commits from
you and Giuseppe would come through.

So for the time being, these seeds contain just the mc triplets, except
instead of using the MC initial state as input to the KF fit over the
seeds, we now use the conformal fit as the initial state for the KF fit
over the seeds.  So the triplet hits themselves are exactly the same as
before.

As the CF is not perfect, given such a small lever arm, it can sometimes
give a crappy initial estimate that leads to tracks getting lost, most
notably before the first build layer.

Once we include the real seeding, however, the story will be the same, as
the seeding can produce at most one triplet that exactly matches mc.  So,
there will be lots of triplets with 2/3 and 1/3 (and some with 0/3) hits
matching a single mc track.  As I have shown before, these fake seeds tend
to die out early, while the real triplets go as far as they would in the
ideal case being tested here.

Let me know if you have further questions.

Peace,
Kevin

dan131riley:
@kmcdermo,

Some (ok, a bunch of) nits.

This first might be controversial, but I'd prefer using std::sqrt, std::abs, etc. everywhere.  It's true that for most of our code sqrtf (for example) is better than sqrt, since sqrt is double--and we have the double version in a lot of places where we want single, so we're doing more double precision math than we need to.  But std::sqrt is usually better because it automatically matches the type of the argument...except of course there's a problem with stuff like

float px  = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);

since 1. is a double, that needs to be 1.f and -1.f to get single precision with std::abs.

I still see

  float phi = getPhi(vry,vrx);
  float px  = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);
  float py  = fabs(pt*sin(phi))*((y[1]-y[0])>0. ? 1. : -1.);

I thought you were going to get rid of those, since that can be rewritten as

  const float invvrxy = 1.0f/std::sqrt(vrx*vrx + vry*vry);
  const float cosphi = vry*invvrxy;
  const float sinphi = vrx*invvrxy;
  float px = std::abs(pt*cosphi)*((x[1]-x[0])>0.f ? 1.f : -1.f);
  float py = std::abs(pt*sinphi)*((y[1]-y[0])>0.f ? 1.f : -1.f);

and you really don't need phi at all (I sent you code for that Tuesday last week).  I see the Matriplex version also has the same extra trig ops.

Next, in MkFitter::ConformalFitTracks(), another thing we should be careful about, code like

    float px, py, pt, pz, phi, theta;
    px = Par[iC].ConstAt(n, 3, 0); py = Par[iC].ConstAt(n, 4, 0);
    pz = Par[iC].ConstAt(n, 5, 0); pt = hipo(px,py);
    theta = getTheta(pt,pz);
    phi   = getPhi(px,py);

where those are temps that are only set once should instead be written with "const float" like so:

    const float px = Par[iC].ConstAt(n, 3, 0);
    const float py = Par[iC].ConstAt(n, 4, 0);
    const float pz = Par[iC].ConstAt(n, 5, 0);
    const float pt = hipo(px,py);
    const float theta = getTheta(pt,pz);
    const float phi   = getPhi(px,py);

Telling the compiler that those values aren't going to be changed gives it more opportunities for optimization (you might think the compiler would be able to figure that out for itself, but from what I've read in the Intel optimization guides it doesn't always, partly because it's very conservative about anything that might have its address taken anywhere--that also shows up in some of the places vectorization needs hints).

And it does look you've got more Matriplex's than you need, for example in

  MPlexQF vrx, vry, phi, pT, pT2, px, py, pz, pz2;

vrx, vry and phi don't look like they are ever used outside that one loop, so they could just be scalar const floats inside the loop (and, as previously mentioned, you don't actually need phi at all).  pT2 and pz2 might not be worth it, looking at the optimization reports might help decide whether the extra matriplexs are worth avoiding a few multiplies.

-dan

dan131riley:
@kmcdermo,

One more thing I just thought of--I haven't tested this, but I believe something like

float px = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);

can be written much more nicely (and efficiently) as

float px = std::copysign(pt*cos(phi), x[1]-x[0]);

There's a bunch of those in both the SMatrix and Matriplex versions.

I'll probably think of something else in a minute or two...

-dan

kmcdermo:
>
>
> One more thing I just thought of--I haven't tested this, but I believe
> something like
>
> float px = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);
>
> can be written much more nicely (and efficiently) as
>
> float px = std::copysign(pt*cos(phi), x[1]-x[0]);
>
> There's a bunch of those in both the SMatrix and Matriplex versions.
>
> I'll probably think of something else in a minute or two...
>
>
This is easy enough to check, I will test this now locally.

kmcdermo:
comments are good!

This first might be controversial, but I'd prefer using std::sqrt,
> std::abs, etc. everywhere. It's true that for most of our code sqrtf (for
> example) is better than sqrt, since sqrt is double--and we have the double
> version in a lot of places where we want single, so we're doing more double
> precision math than we need to. But std::sqrt is usually better because it
> automatically matches the type of the argument...except of course there's a
> problem with stuff like
>

I leave this decision to Matevž and you.  I was told to switch std::sqrt
everywhere to sqrtf sometime back, so I did.  I agree using just single is
better.


> float px = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);
>
> since 1. is a double, that needs to be 1.f and -1.f to get single
> precision with std::abs.
>

Okay, but as you say, this can be resolved once we decide on std::abs vs
fabs, std::sqrt vs sqrtf


> I still see
>
> float phi = getPhi(vry,vrx);
> float px = fabs(pt*cos(phi))*((x[1]-x[0])>0. ? 1. : -1.);
> float py = fabs(pt*sin(phi))*((y[1]-y[0])>0. ? 1. : -1.);
>
> I thought you were going to get rid of those, since that can be rewritten
> as
>
> const float invvrxy = 1.0f/std::sqrt(vrx*vrx + vry*vry);
> const float cosphi = vry*invvrxy;
> const float sinphi = vrx*invvrxy;
> float px = std::abs(pt*cosphi)*((x[1]-x[0])>0.f ? 1.f : -1.f);
> float py = std::abs(pt*sinphi)*((y[1]-y[0])>0.f ? 1.f : -1.f);


Ah, I did this in the error propagation, but not here... okay.  I will fix
this and test.

>
>
and you really don't need phi at all (I sent you code for that Tuesday last
> week). I see the Matriplex version also has the same extra trig ops.
>
> Next, in MkFitter::ConformalFitTracks(), another thing we should be
> careful about, code like
>
> float px, py, pt, pz, phi, theta;
> px = Par[iC].ConstAt(n, 3, 0); py = Par[iC].ConstAt(n, 4, 0);
> pz = Par[iC].ConstAt(n, 5, 0); pt = hipo(px,py);
> theta = getTheta(pt,pz);
> phi = getPhi(px,py);
>
> where those are temps that are only set once should instead be written
> with "const float" like so:
>
> const float px = Par[iC].ConstAt(n, 3, 0);
> const float py = Par[iC].ConstAt(n, 4, 0);
> const float pz = Par[iC].ConstAt(n, 5, 0);
> const float pt = hipo(px,py);
> const float theta = getTheta(pt,pz);
> const float phi = getPhi(px,py);
>
>
easy enough fix, will change, test, and commit.


>
> And it does look you've got more Matriplex's than you need, for example in
>
> MPlexQF vrx, vry, phi, pT, pT2, px, py, pz, pz2;
>
> vrx, vry and phi don't look like they are ever used outside that one loop,
> so they could just be scalar const floats inside the loop (and, as
> previously mentioned, you don't actually need phi at all). pT2 and pz2
> might not be worth it, looking at the optimization reports might help
> decide whether the extra matriplexs are worth avoiding a few multiplies.
>
> Okay, so being my first time writing this kind of code, I just went nuts
with MPlexes... which is why I had that silly comment in the code for "do I
really need all these temp mplexes???"

And this is where I was confused... scalar const floats vs MPlexes.
Because then in principle, I can just replace everything after the actual
matrix multiplication with scalar consts floats, and make it one big simd
loop, no?

i.e., this can just be all scalar float consts with just one big #pragma
simd:
https://github.com/kmcdermo/mictest/blob/cfmplex_port/mkFit/ConformalUtilsMPlex.cc#L109-L222

Peace,
Kevin

kmcdermo:
@dan131riley should have been more careful.

MkFitter::ConformalFitTracks(), the scalar float temps you mention are all inside a #DEBUG portion of the code, so I don't think it really matters what happens here.  I ran with the debug on a few times, and it ran vectorized (i.e. spitting out things as if the loop went in steps of 8).  

But okay, seeing as this is a minimal change, and does not really affect the overall code, making it const is fine.

cerati:
after discussing with @kmcdermo better to merge this one now and add improvements in a future PR when ready

PR #46, opened 2016-05-12
Change mom coord

cerati:
This PR switches the default momentum coordinates in the Track object from global cartesian to quasi-polar (1/pT,phi,theta). In this way we do not need to convert back and forth from the two coordinate systems in the KalmanUtils.
The old behaviour can still be recovered disabling #define POLCOORD in Config.h.

What still needs to be ported to the new coordinates are the SMatrix version of Propagation and of KalmanUtils. Also, the cuda version is not changed for the moment.

Also, a few deprecated functions are removed in this PR.

The timing performance is improved and goes back to the version before PR39, see plots at:
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-11-PR46/


srlantz:
@cerati,

For some time now we have been stumbling over what to call your new coordinates. Now that we are about to adopt them for tracking, I propose calling them "curvature coordinates". 

- Reasons against the old label: the coords are scarcely "polar", or even quasi-polar. First, they apply to the momentum rather than the position; second, the "distance" is actually 1/magnitude; third, they are neither spherical-polar nor cylindrical-polar, but somewhere in between.

- Reasons in favor of the new label: "curvature" is a much better description of what these coords actually correspond to, I believe. 1/pT is directly related to track curvature. And the angles are (I think) a way of describing the spatial direction of curvature associated with the given position parameters. (Please correct me if this interpretation is lacking somehow!)

If you agree, then it may be a good idea to change the #define POLCOORD to, say, CURVCOORD, so we don't get stuck in the current terminology (as I don't think anyone is totally satisified with it).

I will be interested to get your reaction...

dan131riley:
My two cents...

Position vs. momentum is a red herring--a coordinate system is a coordinate system, whether it's applied to position space or momentum space or some completely abstract space.  Polar is definitely wrong, since polar is a 2-d coordinate system.  Spherical is closer, except curvature doesn't add like a distance.  In physics there's the "wave vector", with length proportional to the inverse of the wavelength, and there are spherical wave transforms that correspond to the conformal (angle preserving) group of transformations by the reciprocal of the radius, which suggests "spherical wave coordinates" (or maybe "wave spherical").

srlantz:
@dan131riley ,

Thanks for your input (worth more than 2 cents surely). Technically your terminology is reasonable, but "curvature coordinates" may convey a better intuitive sense of what these coordinates are trying to achieve. I am interested in Giuseppe's reaction.

In defense of my previous point #1 about polar... probably what I should have said was, "they apply to the momentum BUT NOT the position." They are kind of semi-polar, in other words. Nothing wrong with that. But "polar" without qualifiers brings to mind primarily a spatial coordinate system, with other vectors being defined in reference to that system (at least to my mind).

cerati:
what about CCS coordinates, from cartesian-curvature-spherical?
One could add also G6 (from global, dimension 6) but then it would become
difficult to pronounce...

srlantz:
We can talk about it tomorrow if people want. My point is simply that the goal of these "coordinates" is to describe a geometrical object, namely a helix. We are saying that the current best estimate of the track is a z-oriented helix with pitch angle theta and lateral "curvature" 1/pT, and that point x,y,z is a position at azimuth phi (+/- pi/2) on this helix. (OK, to make 1/pT into a real curvature, the charge and B should be included to make it 1/gyroradius; not sure if I have omitted a relativistic gamma in some of the definitions.) The new coordinates just make it a bit clearer that we have been talking about such a geometrical object all along. So call them "helix/helical parameters"?

cerati:
I do not think we should devote part of the meeting to the naming
convention, our 1h slot is too short already.
Just to say that "helix parameters" is a too generic definition that
applies to any parametrization (cartesian, "polar", curvilinear, perigee,
etc.).

Our parametrization has two components:
- a global cartesian position
- a spherical momentum, modified so that the magnitude of the 3D radius
vector is replaced with 1./pT, where 1./pT is the magnitude of the 2D
radius vector and is proportional to the curvature (the fact that pT is 2D
while R is 3D is why I initially suggested polar rather than spherical)

Any name capturing these features is fine with me.

cerati:
merging, we can update the name later...

dan131riley:
CCS (cartesian-curvature-spherical) is fine by me, and I agree about
not using meeting time for this, too close to bikeshedding.

-dan

srlantz:
Agreed, today's meeting will be full enough. But I disagree with the "bikeshedding" characterization. To me, names are important, and they should correctly reflect the concepts behind them.

Steve

srlantz:
This is coming too late for PR #46, but I too am fine with the label CCSCOORD for "Cartesian-Curvature-Spherical".

Or maybe... CCDCOORD, with D=direction? The angles indicate the instantaneous direction of motion of the charged particle following the helix, right?  I maintain that the "spherical" angles also have a clear interpretation as two directional parameters of the helix itself: theta = arctan(pT/pz) = pitch angle, or helix angle; phi = arctan(py/px) = azimuth on the helix, rotated by pi/2. (But maybe "CCD" confuses the experimentalists.)

Steve

PS. Thanks to Dan for reminding me that I hadn't yet responded to Giuseppe's "CCS" proposal. Thanks also to him for chipping in one more argument as to why "helical parameters" is not the best match.

PR #48, opened 2016-05-13
switching input method for the fit test from CopyIn to SlurpIn…

srlantz:
...MatriplexSym::Subtract now vectorizes
1. Code performs best when MPLEX_USE_INTRINSICS is defined
2. Testing has not been performed with USE_CUDA, it may well be broken in SlurpInTracksAndHits

### DIFF
--- a/Matriplex/Matriplex.h
+++ b/Matriplex/Matriplex.h
@@ -125,6 +125,8 @@ public:
    {
       for (int i = 0; i < kSize; ++i)
       {
+        // Next loop vectorizes with "#pragma ivdep", but it runs slower
+	// #pragma ivdep
         for (int j = 0; j < N; ++j)
         {
            fArray[i*N + j] = * (const T*) (arr + i*sizeof(T) + vi[j]);
--- a/Matriplex/MatriplexSym.h
+++ b/Matriplex/MatriplexSym.h
@@ -178,6 +178,7 @@ public:
    {
       // Does *this = a - b;
 
+#pragma ivdep
       for (idx_t i = 0; i < kTotSize; ++i)
       {
          fArray[i] = a.fArray[i] - b.fArray[i];
--- a/mkFit/MkFitter.cc
+++ b/mkFit/MkFitter.cc
@@ -81,6 +81,85 @@ void MkFitter::InputTracksAndHits(std::vector<Track>&  tracks,
   }
 }
 
+void MkFitter::SlurpInTracksAndHits(std::vector<Track>&  tracks,
+                                  std::vector<HitVec>& layerHits,
+                                  int beg, int end)
+{
+  // Assign track parameters to initial state and copy hit values in.
+
+  // This might not be true for the last chunk!
+  // assert(end - beg == NN);
+
+  Track &trk = tracks[beg];
+  const char *varr       = (char*) &trk;
+  const int   off_error  = (char*) trk.errors().Array() - varr;
+  const int   off_param  = (char*) trk.parameters().Array() - varr;
+
+  int idx[NN]      __attribute__((aligned(64)));
+  int itrack;
+
+#ifdef USE_CUDA
+  // This openmp loop brings some performances when using
+  // a single thread to fit all events.
+  // However, it is more advantageous to use the threads to
+  // parallelize over Events.
+  omp_set_num_threads(Config::numThreadsReorg);
+#pragma omp parallel for private(itrack)
+#endif
+  for (int i = beg; i < end; ++i) {
+    itrack = i - beg;
+    Track &trk = tracks[i];
+
+    Label(itrack, 0, 0) = trk.label();
+
+    idx[itrack] = (char*) &trk - varr;
+
+    Chg(itrack, 0, 0) = trk.charge();
+    Chi2(itrack, 0, 0) = trk.chi2();
+  }
+
+#ifdef MIC_INTRINSICS
+  __m512i vi      = _mm512_load_epi32(idx);
+  Err[iC].SlurpIn(varr + off_error, vi);
+  Par[iC].SlurpIn(varr + off_param, vi);
+#else
+  Err[iC].SlurpIn(varr + off_error, idx);
+  Par[iC].SlurpIn(varr + off_param, idx);
+#endif
+  
+// CopyIn seems fast enough, but indirections are quite slow.
+// For GPU computations, it has been moved in between kernels
+// in an attempt to overlap CPU and GPU computations.
+#ifndef USE_CUDA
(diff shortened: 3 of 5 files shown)


cerati:
Benchmarking results: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-13-PR48/
All good, especially this one (as expected):
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-13-PR48/mic_vu_fitspeedup.png
which is much improved on top of:
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-13-PR47/mic_vu_fitspeedup.png
Merging.

cerati:
3 PR merged, what a glorious day :-)

dan131riley:
Here's what I'm getting for my next PR (with 48 merged):

http://www.lepp.cornell.edu/~dsr/mic-track/2016-05-13-std-math/mic_vu_fitspeedup.png

That's all from avoiding doubles and using "const" to minimize temporaries.

-dan

kmcdermo:
whoa! that's totally awesome.

kmcdermo:
Just poking around the other benchmarks, it seems that some have regressed,
e.g.:

http://www.lepp.cornell.edu/~dsr/mic-track/2016-05-13-std-math/mic_th_speedup.png

and this also seems to not achieve the 2x speedup we saw before for CEST:

http://www.lepp.cornell.edu/~dsr/mic-track/2016-05-13-std-math/mic_vu_time.png

dan131riley:
> On May 14, 2016, at 14:39, Kevin McDermott <notifications@github.com> wrote:
> 
> Just poking around the other benchmarks, it seems that some have regressed,
> e.g.:
> 
> http://www.lepp.cornell.edu/~dsr/mic-track/2016-05-13-std-math/mic_th_speedup.png

That one is just noisy--it moves around a lot from run to run, especially at high
thread counts (> 120).

> and this also seems to not achieve the 2x speedup we saw before for CEST:
> 
> http://www.lepp.cornell.edu/~dsr/mic-track/2016-05-13-std-math/mic_vu_time.png

That's still slightly better than PR#48, so there's no recent regression.  Before
that, the history is complicated by the effects of the coordinate changes.

-dan

dan131riley:
I was curious about the cause of the significant performance gain I saw, so I went through the (many, many small) changes to isolate the cause (and convince myself it was real), and the answer was a big surprise to me--it wasn't at all what I would have guessed.

Adding const, being careful about inadvertent doubles, and moving temporaries inside loops all made little difference.

The actual culprit is in mkFit/PropagationMPlex.cc, specifically lines like

      const float& xin   = inPar.ConstAt(n, 0, 0);
      const float& yin   = inPar.ConstAt(n, 1, 0);
      const float& zin   = inPar.ConstAt(n, 2, 0);
      const float& ipt   = inPar.ConstAt(n, 3, 0);
      const float& phiin = inPar.ConstAt(n, 4, 0);
      const float& theta = inPar.ConstAt(n, 5, 0);

and the improvement came from changing the "const float&" to "const float".

The reason, apparently, is that while inPar is aligned, the individual elements aren't, so using a reference forces the compiler into unaligned accesses.  In optrpt (turned up to level 5), the version with the references has a bunch of

   remark #15389: vectorization support: reference xin has unaligned access   [ PropagationMPlex.cc(575,18) ]
   remark #15389: vectorization support: reference yin has unaligned access   [ PropagationMPlex.cc(575,18) ]
   remark #15389: vectorization support: reference pxin has unaligned access  [ PropagationMPlex.cc(586,22) ]

that aren't in the "const float" version, followed by

   remark #15415: vectorization support: gather was generated for the variable pxin: masked, strided by 1   [ PropagationMPlex.cc(585,22) ]
   remark #15415: vectorization support: gather was generated for the variable pyin: masked, strided by 1   [ PropagationMPlex.cc(585,32) ]
   remark #15415: vectorization support: gather was generated for the variable pzin: masked, strided by 1   [ PropagationMPlex.cc(592,22) ]

and when we get to the vectorization report, the version with references has

   remark #15301: SIMD LOOP WAS VECTORIZED
   remark #15450: unmasked unaligned unit stride loads: 3 
   remark #15456: masked unaligned unit stride loads: 39 
   remark #15462: unmasked indexed (or gather) loads: 14 
   remark #15475: --- begin vector loop cost summary ---
   remark #15476: scalar loop cost: 679 
   remark #15477: vector loop cost: 361.370 
   remark #15478: estimated potential speedup: 1.620 
   remark #15487: type converts: 11 
   remark #15488: --- end vector loop cost summary ---

while removing the references (with no other changes) gives

   remark #15301: SIMD LOOP WAS VECTORIZED
   remark #15475: --- begin vector loop cost summary ---
   remark #15476: scalar loop cost: 649 
   remark #15477: vector loop cost: 258.310 
   remark #15478: estimated potential speedup: 2.070 
   remark #15487: type converts: 11 
   remark #15488: --- end vector loop cost summary ---

The compiler does end up allocating a bunch more stack space for the temporaries, but it can make sure they're properly aligned, and that gives a significant speedup in the propagation loop vectorization.

If I then add on all the other changes (const, avoiding doubles), we get to

   remark #15301: SIMD LOOP WAS VECTORIZED
   remark #15475: --- begin vector loop cost summary ---
   remark #15476: scalar loop cost: 582 
   remark #15477: vector loop cost: 247.810 
   remark #15478: estimated potential speedup: 1.920 
   remark #15487: type converts: 1 
   remark #15488: --- end vector loop cost summary ---

where the type conversions are down significantly (I don't know why there's still 1, there aren't any reported), and the loop is a few percent faster, but the vector speedup is actually a little less because the scalar loop cost went down a lot more than the vector cost.

I changed the "float&" references to "float" variables because I suspected the compiler could do better managing the lifetimes of temporaries instead of always going back to the source parameter inPar, but I was not expecting anything this dramatic.  I'm going to have to reconsider how I think about temporaries, and how they interact with vectorization.

-dan

kmcdermo:
This is very interesting.  Do we have other places in the optrpts that point to similar unaligned accesses?  

Can we say we aren't really hitting the cache limit then if the tradeoff between unaligned accesses and temporaries is this dramatic?

osschar:
> Can we say we aren't really hitting the cache limit then if the tradeoff between
> unaligned accesses and temporaries is this dramatic?

You can see cache / memory b/w limitations when we run two threads on the same 
core. So if you want to see impact on this, compare slopes on mic threading plot 
between 60 and 120 threads.

Temporaries don't necessarily impact cache ... especially if they are short 
lived enough to just stay in the registers. With two threads per core pressure 
on registers increases as well.

PR #49, opened 2016-05-17
Reduce type conversions, add const, help the compiler with better hints

dan131riley:
A large number of small changes.

* change calls to cos, log, abs, etc. to use std:: versions that polymorphically adjust to the argument types
* add const wherever possible
* avoid implicit conversion to double
* reduce the scope of temporaries
* let Debug.h be included multiple times until DEBUG is defined
* implement the optimizations I previously suggested for ConformalUtilsMPlex.cc

No changes in algorithms or functionality.  Most changes have a few percent effect, but favoring temporaries over references in PropagationMPlex.cc actually gives a quite significant improvement in vectorization on Xeon Phi (with a smaller improvement on SB) due to alignment constraints.

Supporting material in phiphi:/home/dsr/mictest-mt/2016-05-16-std-math

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -26,6 +26,8 @@ namespace Config
   bool  useCMSGeom = false;
   bool  readCmsswSeeds = false;
 
-  bool  super_debug = false;
   bool  cf_seeding  = false;
+  bool  cf_fitting  = false;
+
+  bool  super_debug = false;
 }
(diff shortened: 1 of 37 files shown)


cerati:
Thanks @dan131riley !
Plots are copied to: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-05-16-PR49/
and the PR is being merged.

PR #50, opened 2016-06-03
Remove usage of overlapping eta partitions for hit grouping

osschar:
* Introduce z-binning and internal phi-bin sorting into HitStructures::LayerOfHits.

* Sorting is done using RadixSort, implementation in mkFit/Ice.

* Modify all track finding functions to use LayerOfHits.

* Use N_proc parameter in MkFitter::AddBestHit() as well as in all mkFit Propagation and KalmanUtils functions.

* Remove obsolete functions in MkFitter dealing with vector<Hit>.

* Remove HitStructures::BunchOfHits and all functions for their usage and handling.

* Consolidate debug mode, remove obsolete g_debug global.

* Calculate some Config constants that depend on command-line parameter --num-tracks after parsing of command-line options.

* Add support for using gcc-5 from MacPorts on OSX.

Performance notes
=================

Tuning done for selection of hits to be considered for each candidate is such
that the code runs a bit faster than before (5%) and gives a bit better
results than before (1.5% more good tracks).

Opening up of phi windows for one extra phi bin on each side (1/1024 2Pi)
gives further 1.5% more of good tracks and takes 10% of runtime.

Further speedup is possible once we understand errors and minimal phi and z
windows as the new code make it easy to do hit pre-selection based on dphi /
dz between the hit and predicted track intersection.

The above speedups are for tbb, cloner-same-thread, n_thr = 1.

There is no change in seed processing ... they are processed in the same order
/ loop structure as if the eta bins still existed. Chaning this might improve
work balancing with multi-threading.

Notes on discrepancies between MIC / Host
=========================================

This code has some discrepancies between number of hits found on MIC /
host. Most of them seem to originate from cumulative effect of
rounding during the initial fit of seeds. Giuseppe and Matevz decided
to postpone further investigation. See comments starting with DDDD in
MkBuilder.cc to see how to reproduce the problem and do full debugging.

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -1,9 +1,13 @@
-// Include of Config.h not yet required
+#include "Config.h"
 
 namespace Config
 {
   int nTracks = 20000;
   int nEvents = 10;
+
+  // Dependent constants, assigned after processing of commandline options
+  int maxHitsPerBunch;
+  int maxCandsPerEtaBin;
 
   // Multi threading and Clone engine configuration
   int   numThreadsFinder = 1;
@@ -30,4 +34,10 @@ namespace Config
   bool  cf_fitting  = false;
 
   bool  super_debug = false;
+
+  void RecalculateDependentConstants()
+  {
+    maxCandsPerEtaBin = std::max(100, maxCandsPerSeed * nTracks / nEtaPart);
+    maxHitsPerBunch   = std::max(100, nTracks * 12 / 10 / nEtaPart) + maxHitsConsidered;
+  }
 }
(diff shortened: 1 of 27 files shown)


osschar:
Benchmarking results: http://uaf-2.t2.ucsd.edu/~matevz/mictest/2016-06-02-mt-nix-eta-bins/

osschar:
And the latest benchmarks: http://uaf-2.t2.ucsd.edu/~matevz/mictest/2016-06-07-mt-nix-eta-bins-pr/

This restores performance for NN=2.

cerati:
Rerun benchmarking: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-06-08-PR50/
All good, merging.

PR #51, opened 2016-06-03
Cmssw tests collisions

cerati:
Minor changes for testing on CMSSW simulation of collision events, used to get these results:
https://dl.dropboxusercontent.com/u/13022849/compareCMSSW-noFWD.pdf

Mostly independent from PR50, but small conflict in printouts. 
There is no real change in default, benchmarking not run for now.


cerati:
Update.
Benchmarking run, results unchanged: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-06-09-PR51/
Commits from yesterday are after merging PR50 and were not included for the results summarized in the slides above.
Merging.

srlantz:
Giuseppe,

I think one change crept into PR51 unnoticed, namely, the Makefile.config specifies USE_INTRINSICS := -DMPLEX_USE_INTRINSICS, whereas previously this line was commented out. It is not a major effect, but the change does affect comparisons with benchmarks prior to PR51.

It also affects VTune results. It seems the "loads blocked by store forwarding" issue I reported on phiphi's SNB host (using the PR56 code) can be traced partly to the FMA function defined by Matevz in MatriplexCommon.h. This FMA just calls the _mm256_mm_ps() intrinsic. I suspect that previously, this particular issue didn't show up so much because we disabled intrinsics by default

Looking forward to PR56: VTune still says that "loads blocked by store forwarding" is an issue, even if I revert to the previous default and turn off the intrinsics. However, it seems the impact is much less with the intrinsics disabled.

Steve

osschar:
On MIC we use the fma intrinsic:
`#define FMA(a, b, v)  _mm512_fmadd_ps(a, b, v)`
and on AVX:
`    inline __m256 FMA(const __m256 &a, const __m256 &b, const __m256 &v)
    {
      __m256 temp = _mm256_mul_ps(a, b); return _mm256_add_ps(temp, v);
    }
`
so it adds a temporary and uses an extra register. If you look at code generated by GenMul you'll see we can have several fmas in flight, probably up to four.

I think avx2 already has fma ... as does avx-512.

cerati:
the change in Makefile.config actually comes from merging PR50, so its effect should be visible already there, can you confirm?
In any case it should be reverted.

srlantz:
@cerati,

Huh, you're right--this same change also appears in PR50. I thought it would be sufficient to look back at "Files changed" for each PR until I found the most recent occurrence of this change in Makefile.config. But I guess the "Files changed" tab does not show a comparison with the numerically preceding PR that was merged. It must show a comparison with whatever was the last pull by the creator of the PR, which might reflect an earlier state of the repository. Otherwise, if this line was changed in PR50, how can it then be changed again in exactly the same way, in PR51?

cerati:
Hi Steve, this is because I merged PR50 in my branch, and when you merge
the commits from the other branch get included and show up also in the PR.

PR #54, opened 2016-06-14
use correct error in estimation of dphi

kmcdermo:
small bug noticed in selecting the hit indices while building -- use err(0,1) for estimation of dphi.


kmcdermo:
benchmarking plots are here:  https://kmcdermo.web.cern.ch/kmcdermo/pr54/

Comparing to last benchmarking plots (PR51), nHits improves from 9.868 to 9.964, as one would expect (9.964!!! pretty impressive.  hump at 5<= nHits <= 7 is now gone).

Speedup is the same within fluctuations.  Timing is about 3-5% slower for both parallelization and vectorization (which is not unexpected, given we recover 1% more hits).

cerati:
thanks for spotting this!
plots copied to http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-06-14-PR54/
merging.

PR #55, opened 2016-06-15
cmssw benahcmarking tests

cerati:
Add build test of CMSSW TTbarPU35 to benchmarking scripts.


cerati:
Plots: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-06-15-PR55/

kmcdermo:
@cerati cool!  Out of curiosity, how much longer does the benchmarking take with the cmssw building tests?

cerati:
It took ~2h, I think it is not much different from before

cerati:
some slides, just for reference (no need to present them):
https://dl.dropboxusercontent.com/u/13022849/compareCMSSW-noFWD-update.pdf

PR #56, opened 2016-06-20
tbb fitter, adaptive seeds per task, and some miscellany

dan131riley:
Benchmarks are in phiphi:/home/dsr/mictest-mt/2016-06-20-tbb-fitter

This converts runFittingTestPlex() to TBB threading; this does actually result in some modest scalability improvements, particularly at high thread counts.  I also discovered that SlurpInTracksAndHits() is only safe for full matriplexes, so this version calls InputTracksAndHits() instead for partial matriplexes, replacing a min() operation.  Also pulled the Pool struct out into a separate header so that the fitter could use it, and moved the initialization of the ExecutionContext that holds fitters and cloners for TBB out of the constructor.  In the constructor, it was being called before the options had been processed, and so was initializing for the wrong thread count, and always creating one CandCloner thread.  Initialization now follows the same pattern as the OMP versions.

Added an adaptive calculation for seeds per task, which calculates a stride based on the number of seeds in the current bin and the thread count, using that value if it is smaller than the configured stride.  This substantially improves the performance of the TBB builder for the CMSSW tests with high thread counts.

The fitter now always calls OutputFittedTracks(), and make_validation_tree() has been modified to print a count of tracks better than 80% and the mean and standard deviation of the pulls, based on code Matthieu wrote for the GPU version.  Also updated make_validation_tree() to be independent of coordinate system.

maxCandsPerEtaBin was being overrun for small track counts (hence larger variance).  A safety factor was added that increases the value the most for small track counts.

hi.zbin was also being overrun, likely due to smearing pushing some hits off the end of the detector.  Added a check for this.  This change resulted in a very small improvement in the hits found, and changed the platform and vector-length differences, so running off the end of that vector was one contributor to those differences.

benchmark-mic-cmssw.sh wasn't found, benchmark-cmssw-mic.sh was, so updated runBenchmark.sh, @cerati should check this is correct.  Also some other miscellaneous bug fixes for Debug and cartesian coords that aren't tested in the benchmarks, and a correction to a mistaken change I made to the time reporting (does not affect benchmarks).


cerati:
plots copied at:
http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-06-20-PR56/

cerati:
@dan131riley yes you are correct about benchmark-cmssw-mic.sh (I ran it by hand, and later added the buggy line to the main script)

PR #57, opened 2016-07-21
Real seeding in mkFit + TTreeValidation in mkFit

kmcdermo:
This is a mega merge: I have incorporated triplet seeding into mkFit as well as the TTreeValidation in mkFit.  The main advantage of the TTreeValidation is that it can do efficiencies, fake rates, duplicate rates, nHit matching, etc. vs pT, eta, phi, etc.  

This PR also includes a number of fixes, namely the Conformal fit as well as a mapping between m_event->layerHits_[ilayer] and m_event_of_hits.m_layers_of_hits[ilayer].m_hits.  This was necessary to have a unified seed fit as well as restore the MC matching between reco and sim based on the 75% hit matching criterion.  I also added a few printouts in the benchmarking scripts just echoing where you are in the various loops.  

As always, here are the benchmarking plots: https://kmcdermo.web.cern.ch/kmcdermo/pr57/benchmarks/

One thing to note: in trying to track down the differing nHits/track between this PR and PR#56, I noticed I could not even get the nHits/track to be constant across the same configuration running it multiple times with a vanilla checkout of devel.  The first differences came from some bugs that I have iterated on with G.C., M.T., D.R., et. al.  The residual differences of the nHits differences come from the fact that the simulation is not one to one.

Reading the same simulation file between vanilla devel with PR#56 and this PR, the output nHits is the same (with nTh = 1, nVU = 1, CEST, host). 

-----------------------------------------------------

I have included two examples of using the TTreeValidation (with the ifdef WITH_ROOT enabled): 
CEST (all other config standard, i.e. 20k tracks/event * 10 events): 
1) ./mkFit --build-ce --cloner-single-thread --normal-val
Results: https://kmcdermo.web.cern.ch/kmcdermo/pr57/vanilla/

Read+Write "Real Seeding" Best Hit:
2a) ./mkFit --write --num-tracks 500 --num-events 100 --normal-val
2b) ./mkFit --read --find-seeds --build-bh  --normal-val
Results: https://kmcdermo.web.cern.ch/kmcdermo/pr57/shoots/
("shoots", like seeds, roots and shoots...)

The text files "totals*.txt" show in plain text the fake rates, duplicate rates, and efficiencies of the different track collections.  The flag --normal-val indicates use of the ROOT TTreeValidation.  In order to do this while reading in a binary file of tracks, however, we need to save the simHits info, which is why the flag also appears in 2a.  If you are interested in producing the validation plots after running the building, simply do: root -l -b -q runValidation.C

We also can do the TTreeValidation with CMSSW tracks as well, although we would have to change the runValidation plots to go up to greater than Config::nTotHit.


cerati:
copied to: http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-07-21-PR57/
I let @osschar and @dan131riley deal with merging

osschar:
+1 @dan131riley please sign off

kmcdermo:
I added directly a one-line change (insertion of a mutex) in TTreeValidation.cc.  After seeing some segfaults, I realized I was writing to the same object (map of track states) across all threads when doing multithreaded simulation + the full TTree validation.  Tested locally and no longer segfaults.

dan131riley:
+1, will merge shortly, couple of really minor style points...

This is kinda overkill since constexpr is evaluated at compile time:

-  constexpr float PIOver2  = Config::PI / 2.0;
+  constexpr float PIOver2  = Config::PI / 2.0f;

And here:

+    outErr.At(n, 0, 0) = x.ConstAt(n, 0, 0)*x.ConstAt(n, 0, 0)*invvarR2 + y.ConstAt(n, 0, 0)*y.ConstAt(n, 0, 0)*varPhi;

dunno how anyone else feels, but I prefer the "at-less" style:

    outErr(n, 0, 0) = x(n, 0, 0)*x(n, 0, 0)*invvarR2 + y(n, 0, 0)*y(n, 0, 0)*varPhi;

I added a const overload to operator() specifically so we didn't need ConstAt().

-dan

osschar:
I also noticed the 2.0f but didn't want to fuss about it :)

I agree with with pure op(), it's much easier to read.

kmcdermo:
@dan131riley I suppose you will modify and commit with the overloaded op()?  I don't mind having to wait for #58 if it makes it easier (I can resolve the conflicts).

dan131riley:
Ok, I merged #58; @kmcdermo, would you kindly pull from the head of devel, fix the merge conflicts, and update your PR?  I already reviewed the conflicts and they look straightforward.  Thanks!

kmcdermo:
Hi @dan131riley, I managed to fix the conflicts, and was trying to compile w/o ROOT... but w/ ROOT it hung in two different spots:
1) At scp of main-mic to mic
2) After removing the scp in the Makefile, it just hung on the linking... 

Any idea what is happening?

kmcdermo:
Hmm, maybe the phi cards were just stuck in some weird state.  I tried again after a make clean, and now it all compiles.  Ready to merge.

dan131riley:
What's actually hanging is the compile of main.cc for the host, you're getting fooled about what's hanging because you're compiling more than one thing at a time.

What's really weird is what's going wrong.  main.cc, compiled with

icc -I. -DUSE_MATRIPLEX -DMPLEX_USE_INTRINSICS -std=c++11 -DCHECKSTATEVALID -DLINEARINTERP -DENDTOEND -DETASEG -DINWARDFIT -DGENFLATETA -I/opt/intel/vtune_amplifier_xe_2016.3.0.463186/include -DUSE_VTUNE_PAUSE -DTBB -pthread -m64 -I/home/matevz/root/include -O3 -g -qopt-report -openmp -mavx -c -o main.o main.cc

is taking up absurd amounts of memory (64GB), to the point where the system starts paging madly and gets really slow.  If I add -DNO_ROOT, it compiles.  If I comment out the line

      MCHitInfo::mcHitIDCounter_ = 0;

it compiles.  And if I leave that line in but change

  TTreeValidation val("valtree.root");

to

  Validation val;

so I still have the same headers, I'm just not instantiating the TTreeValidation class, it's fine.  So somehow it's the combination of

  TTreeValidation val("valtree.root");

and

      MCHitInfo::mcHitIDCounter_ = 0;

driving the compiler nuts.  But I don't see why.

-dan

dan131riley:
still don't understand the problem, but try this workaround:

diff --git a/Hit.cc b/Hit.cc
index 6e6fa53..12148ce 100644
--- a/Hit.cc
+++ b/Hit.cc
@@ -1,3 +1,8 @@
 #include "Hit.h"
 
 std::atomic<int> MCHitInfo::mcHitIDCounter_(0);
+
+void MCHitInfo::reset()
+{
+  mcHitIDCounter_ = 0;
+}
diff --git a/Hit.h b/Hit.h
index 3e2bbf7..c44ba1d 100644
--- a/Hit.h
+++ b/Hit.h
@@ -177,6 +177,7 @@ struct MCHitInfo
   
   int mcTrackID() const { return mcTrackID_; } 
   int layer()     const { return layer_; } 
+  static void reset();
   static std::atomic<int> mcHitIDCounter_;
 };
 typedef std::vector<MCHitInfo> MCHitInfoVec;
diff --git a/main.cc b/main.cc
index 378a077..156654c 100644
--- a/main.cc
+++ b/main.cc
@@ -303,7 +303,7 @@ int main(int argc, const char* argv[])
 
     if (Config::endcapTest) {
       //make it standalone for now
-      MCHitInfo::mcHitIDCounter_ = 0;
+      MCHitInfo::reset();
       ev.simHitsInfo_.resize(Config::nTotHit * Config::nTracks);
       fittestEndcap(ev);
       continue;;

kmcdermo:
Hi @dan131riley , so I made the changes and it compiled without a problem.  It is certainly strange, but thanks for the fix.

PR #59, opened 2016-07-28
Seeding with TBB + Fixes to Validation

kmcdermo:
I will run the benchmarking tomorrow, although it should be delta = zero.  

This PR includes a first attempt at seeding with TBB.  The parallelization performance improves by a factor of 2 from 1 thread to 3 threads for 20k tracks, but then regresses beyond that.  The pragma simd also does not vectorize... I need to play with this more.

However, the main reason for this PR is a fix to the validation.  The TrackExtra's were not aligned properly, and on the off chance that a seed is not input into any of the building routines because the posEta is a NaN or Inf, this screws up the validation.  So I added some extra protection to the validation with a realigning of the track extras (which was already done previously in the SMatrix version after a sort on the tracks in phi).

I will update with the benchmarking tomorrow.


osschar:
Do you want to add something else here? How did benchmarking go? I haven't tried it since Giuseppe updated it for endcap.

kmcdermo:
I will not be adding anymore to this branch.  I will be running a number of other tests to try and get the vectorization performance actually working.  Steve and Dan have provided a number of studies and ideas for this latest version, and I will try them out on a new branch.

I have not given the benchmarking a go.  I can do it do, and attach it later tonight.

osschar:
Yeah, please do ... just to make sure nothing creeps in that would be hard to chase down later. I also need to figure out how to manage the benchmarking results :)

kmcdermo:
Ugh.  So I have (most of) the benchmarks here: https://kmcdermo.web.cern.ch/kmcdermo/pr59/

Conspicuously, the endcap building plots are totally empty.  This is not entirely surprising (and something I am working on now quickly), as the validation was prepared for barrel only and not endcap (even just the text dump validation).  

The rest of the plots are within the noise of the tests (cmssw/toymc, mic/host, fit (barrel and endcap), vu/th, speedup/time).

kmcdermo:
Hi all, 
I made the necessary changes to get the endcap tests running again, and now the hyperlink to pr59 benchmarks above show the full results.  You will notice the first point in the endcap TTBST host is totally wacko, throwing off the scaling quite a bit.  I can rerun this point... but the rest honestly looks fine, so let me know what you think.

The endcap mic building vs. VU takes FOREVER... as evidenced by the scale being too small to handle the times (see Giuseppe's plots to see the same effect).

Other tidbits:
In cmssw seeds + endcap, apparently the simTracks collection is not filled such that the index inside ev.simTracks_ equals the mcTrackID() (ie. label() != index inside vector).  To rectify this, I added an "align" function for the simTracks_ when it comes to validation.

I also noticed that we have a number of cmssw seeds in the endcap with label() == -1.  

There are also seeds with a valid label() that also are missing seed hits in the first two layers (i.e. getHitIdx(ilay) == -1!).  We therefore get a number of zero size tracks from the building, so I assign in the validation any reco track with all hitIdxs == -1 to have a mcTrackID() == -2.  

Any reco track that has even a single hit but is unassociated to an MC track by virtue of the 75% matching criterion is assigned an mcTrackID() of -1.

kmcdermo:
I reran the first two points on benchmark-cmmsw-endcap.sh (i.e. nTh=1,3 for endcap building with cmssw seeds).

As you can see, between:
**Giuseppe original:** http://uaf-6.t2.ucsd.edu/~cerati/mictest/2016-07-23-PR58/cmssw_host_endcap_th_time.png
**This PR:** https://kmcdermo.web.cern.ch/kmcdermo/pr59/cmssw_host_endcap_th_time.png

The first point is no longer crazy and the agreement is much better.  However, the agreement is not one-to-one.  I tried rerunning this about 4 times, and this was one of the better ones.  

I then decided to just check out the original branch for the endcap tests with PR58, and the results are here for that same nThread test:
https://kmcdermo.web.cern.ch/kmcdermo/pr59/tmp/

As you can see:
https://kmcdermo.web.cern.ch/kmcdermo/pr59/tmp/cmssw_host_endcap_th_time.png

This is closer to PR58 for some of the points, but not all.  Again, this was done after running the tests 3 or 4 times, as it turned out nTh=1 for TBBST was also totally insane on this plot on the first try!  

Let me know what you think.

kmcdermo:
Also, as an aside, can we switch to make -j 24 (or at least 12)?  We spend a lot of time in compilation in the benchmarking, and I have noticed it goes a bit faster with 24 threads.

mkFit/MkBuilderEndcap.cc
```diff
@@ -130,13 +130,13 @@ void MkBuilderEndcap::begin_event(Event* ev, EventTmp* ev_tmp, const char* build
...
-  if (Config::readCmsswSeeds==false) m_event->seedTracks_.resize(simtracks.size());
+  if (Config::readCmsswSeeds==false) m_event->seedTracks_ = m_event->simTracks_;
 }
 
-inline void MkBuilderEndcap::fit_one_seed_set_endcap(TrackVec& simtracks, int itrack, int end, MkFitter *mkfp)
+inline void MkBuilderEndcap::fit_one_seed_set_endcap(TrackVec& seedtracks, int itrack, int end, MkFitter *mkfp)
 {
-  mkfp->SetNhits(2); //note this is 2 instead of 3 since we ignore PXB1
-  mkfp->InputTracksAndHits(simtracks, m_event->layerHits_, itrack, end);
+  mkfp->SetNhits(Config::nlayers_per_seed); //note this is 2 instead of 3 since we ignore PXB1
```
dan131riley: Should this be Config::nlayers_per_seed-1?

dan131riley:
+1 for me, @osschar ?

kmcdermo:
Okay, I commented out the endcap building tests for the mic.

mkFit/MkBuilderEndcap.cc
kmcdermo: In case people missed this point, I made Config::nlayers_per_seed an extern, with a default = 3 for the barrel: https://github.com/kmcdermo/mictest/blob/seeds-tbb/Config.cc#L28

If the flag "--endcap-test" is used in ./mkFit, then Config::nlayers_per_seed gets set to 2: https://github.com/kmcdermo/mictest/blob/seeds-tbb/mkFit/mkFit.cc#L529-L532

osschar:
OK, +1 ... merging.

PR #61, opened 2016-09-16
Merge gpu best hit

mpbl:
I have regularly squashed commits. I hope there is not too many of them remaining.

Benchmarks were run on PhiPhi,  results are there: [PR 60 benchmark results](https://mpbl.gitlab.io/physics/pr60/).


mkFit/MkFitter.cc
```diff
@@ -257,7 +266,9 @@ void MkFitter::InputTracksAndHitIdx(const std::vector<Track>& tracks,
 
     for (int hi = 0; hi < Nhits; ++hi)
     {
-
+      // MPBL: It does not seem that these values are that dummies
+      //       Not transfering them to the GPU reduces the number of
+      //       nFoundHits in the printouts.
```
dan131riley: "dummy" comment is out-of-date, we should fix that (in a future PR)

mkFit/MkFitter.cc
```diff
@@ -65,7 +65,10 @@ void MkFitter::InputTracksAndHits(const std::vector<Track>&  tracks,
   // assert(end - beg == NN);
 
   int itrack;
-#ifdef USE_CUDA
+
+// FIXME: uncomment when track building is ported to GPU.
+#if 0
```
dan131riley: For future reference: I'd prefer something like "ifdef USE_CUDA_NOTYET" rather than "if 0"

mkFit/fittestMPlex.cc
```diff
@@ -24,6 +25,7 @@
 
 #include <iostream>
 #include <memory>
+#include <mutex>
```
dan131riley: Why was this added?  (There's already an include of mutex inside "ifndef NO_ROOT", and the only use of a mutex is also inside "ifndef NO_ROOT")

BinInfoUtils.h
```diff
@@ -12,25 +12,37 @@ typedef std::pair<int, int>                 BinInfo;
 typedef std::vector<std::vector<BinInfo>>   BinInfoLayerMap;
 typedef std::vector<BinInfoLayerMap>        BinInfoMap;
 
+#ifdef __CUDACC__
+__host__ __device__
```
osschar: We could have a derived define for those ... so we don't need to have ifdefs all over the place.

PR #65, opened 2016-12-08
Restore Standard Building + Converted to TBB (+ more benchmarking changes)

kmcdermo:
This PR is on top of PR #64 . I have restored the standard building (removed in #64), and converted it to TBB so that it can run with all the new benchmarks (+ soon to be Dan's multiple events in flight).

I also made a number of additional changes to the benchmarks.  This includes renaming the logs and arguments passed to the benchmarking scripts to make it more explicit.  The added benefit is now benchmark plots should appear grouped together when viewed on the web (as the names of the .png files explicitly states what is being measured).  

The full benchmarks are here: https://kmcdermo.web.cern.ch/kmcdermo/restored/

Notice that as expected, the standard building returns the same physics results, but slower and less performant as clone engine.  The scaling is rather smooth on SNB.

As for KNC, the benchmarks are a bit jumpy for all tests, but again, this is expected as there is limited work and not enough events to reach a steady state (as all tests do 10k tracks per event, and the time is now the average time over all events -- technically the average time over all events excluding the first).


osschar:
I reviewed this last time as 64 ... so if it's only better than before I'm all for merging :)

kmcdermo:
Ha!  Well, I certainly think it is better than 64, but I will let you and Dan decide.

PR #66, opened 2016-12-12
Support for multiple events in flight

dan131riley:
Support for multiple events in flight, and a bunch of related stuff.  This is a big PR, sorry.

* Multiple events in flight.  Event and MkBuilder objects are pre-allocated and reused.

* Changed Validation initialization to use factory pattern.  With multiple events in flight, each event stream gets its own validator and creates ROOT files with a serial number.

* Event::Reset() added so Event can be reused.  mcHitIDCounter moved into the Event so multiple Events can be simulated at the same time.

* Removed side thread.  Removed CandCloners from EventTmp, removed MkFitters from MkBuilder, since both were only used by the OMP code.  Removed OMP entirely except for CUDA code.

* Changed file format to add length of each event, so the mutex on event reads can be kept minimal.  Event::read_in can read the old format (with only one event in flight), which can be used to convert files from the old to new format, for example (with the appropriate geometry patches applied):
    ./mkFit//mkFit --input-file /data/nfsmic/cerati/cmssw_100xTTbarPU35_polar_split_endcapEta1p7.bin --input-version 0 --file-name /data/nfsmic/dsr/cmssw_100xTTbarPU35_polar_split_endcapEta1p7.bin --cms-geom --cmssw-seeds

* Made the user directory for the benchmark files customizable via an environment variable.  This currently points to my area because I have the data files in the new data format.

* Tried to rationalize a bunch of options in Makefile.config to be more automatic and configurable via environment variables (e.g., OSXGCC5, OSXMPCLANG).

* Mutex on all print statements in event loop.  New "--silent" flag to suppress all printouts (and associated locks).

* Parallelized loop over layers in MkBuilder::begin_event()

* Fixed a bunch of compiler warnings.  These were mostly unrelated to multiple events, but I wanted to get to zero warnings.

Plots are in

https://www.classe.cornell.edu/~dsr/mic-track/PR66/

I have not added any benchmarks for multiple events in flight yet, so these should (and do) look just like #65 plots.

Some of the changes to Event are sure to have broken the CUDA code.  Unfortunately, I can't compile the code to check, it seems to rely on some headers that I don't have, like cub/util_debug.cuh.  It should be reasonably obvious how to fix the CUDA code by looking at the corresponding changes for the Intel CPUs, but it would be nice to get the CUDA code to compile so it can be checked.


mpbl:
When trying to address the last paragraph of your PR, I get a segfault:
```
...
Total event > 1 fit = 0.00000  --- Build  BHMX = 0.59186  STDMX = 4.79929  CEMX = 3.99621
Total event loop time 19.00940 simtracks 200000 seedtracks 0
Segmentation fault
```

The strange thing is that there is no problem on my mac laptop (clang-800 / gcc 6.1.0). I have tried with icc 16.0.0 on phiphi, and icc 16.0.4 / gcc 5.3.1 on another linux machine and get the segfault.

Which compiler version should I use on linux machines?

dan131riley:
You found a bug in a case we don't test for our benchmarking, since our tests all read from a file instead of generating events.  I've checked in a fix, thanks.

PR #206, opened 2019-03-22
Remove duplicates

areinsvo:
First pass at duplicate removal. Current algorithm is to remove the track with the lower score in a pair of tracks satisfying:
fracHitsShared > 0.75, delta phi < 0.1, std::abs(pt2 - pt1)/maxpt < 0.05 and delta eta < 0.2

This can be re-optimized after we fix the efficiency for short tracks.

Plots can be found here. This includes plots with and without duplicate removal, with and without the mtv-like-val option, and plots comparing the performance within CMSSW:
http://areinsvo.web.cern.ch/areinsvo/MkFit/DuplicateRemoval/forPR/

I haven’t tried to optimize it for timing yet; I wanted to get the physics results out to include in my Connecting the Dots presentation.

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -80,6 +80,8 @@ namespace Config
   bool  readSimTrackStates = false;
   bool  inclusiveShorts = false;
   bool  keepHitInfo = false;
+  bool  useHitsForDuplicates = false;
+  bool  removeDuplicates = true;
   bool  tryToSaveSimInfo = false;
   matchOpts cmsswMatchingFW = hitBased;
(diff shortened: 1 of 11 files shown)


makortel:
> we should also apply the duplicate removal by default when running mkFit in CMSSW, but we will wait a few days until the rebase to CMSSW_10_4 is finished.

Apparently we forgot to turn the removal on for CMSSW... https://github.com/cms-sw/cmssw/pull/27895#issuecomment-530958661

PR #227, opened 2019-06-03
safer delete

slava77:
```delete [] x ``` -> ```operator delete [] (x)```
to avoid locks in jemalloc.

This came up in tests with CMSSW.


tresreid:
I have run benchmarks for this. They can be found here: http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_7_3_19_saferdel/
However, I think Slava's repo might have been private since i was not able to access it. I ended up just making the changes by hand instead of pulling this branch. Seeing as though this was only two lines, i don't think that should be a problem.

PR #228, opened 2019-06-10
Fixing lnx scripts

tresreid:
lnx scripts are fixed after merging/rebasing issue.
Benchmarks can be found here: http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_06_07_19/
lnx computers can be enabled from the command line using the "useLNX" option (0 [default] = phi, 1 = lnx, 2 = lnx +phi) in conjunction with the lnxuser option for the lnx username (if needed. defaults to phi username)


tresreid:
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_06_07_19/
Sorry, I think i renamed all the benchmarks i had when i was cleaning up.

PR #229, opened 2019-07-17
Add val option for mtv requiring sim tracks matched to seeds

areinsvo:
Using --mtv-require-seeds, you can now get MTV-like validation but with the additional requirements that the sim tracks are matched to a seed. All of the benchmark and validation plots are here, including the SIMVAL for the extra option and the normal MTV-like validation.
http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/July2019_AddValOpt/

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -93,6 +93,7 @@ namespace Config
   float minFracHitsShared = 0.75;
 
   bool mtvLikeValidation = false;
+  bool mtvRequireSeeds = false;
   int  cmsSelMinLayers = 12;
   int  nMinFoundHits = 10;
 
--- a/Config.h
+++ b/Config.h
@@ -305,6 +305,7 @@ namespace Config
   extern    int maxCandsPerEtaBin;
 
   extern    bool mtvLikeValidation;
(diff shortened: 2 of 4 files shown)


slava77:
it looks like there was a factor of 2 speedup at high MT
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_7_3_19_saferdel/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/1000-icc-19.0.4/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png

or in speedup units
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_7_3_19_saferdel/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_speedup.png
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/1000-icc-19.0.4/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_speedup.png

is the last PR test done with a different number of events?

Issue #233, opened 2019-08-09
establishing new baseline for benchmarks with "performance" scaling governor


srlantz:
We have decided to keep the CPU scaling governor setting as "performance" on phi3; it was changed it from "powersave" to "performance" sometime between July 9 and 15, which is why our multithreaded benchmarks suddenly looked a whole lot better on phi3 (see issue #232). Benchmarking results should be more consistent (and better) this way; therefore, this would be a good change to make to all our test machines.

This implies we will want to do before-and-after benchmark plots on all our test machines so that we are cognizant of the effects of making this hardware-related change globally. I propose that we do so in connection with a PR (not yet opened) where the benchmark scripts have lines added to them to verify the settings of no_turbo and scaling_governor (and possibly others), to see that they are what we expect on each machine where the benchmarks are run. Hopefully this will avoid any confusion over these settings in the future.

Previously we discussed dropping the phi1 and phi2 machines from regular benchmarking routine We could do that now (i.e., ignoring them from now on) or after the proposed PR. @dan131riley thinks that if we want to retain them for benchmarking in the future, we should make them consistent with the other machines; also these changes may have less dramatic effects on the older machines.

srlantz:
Any comments on this issue? I especially want to know which machine(s) we would like to keep for benchmarking purposes--even if we run benchmarks on them all right now to check powersave vs. performance settings (as well as to document the performance of the "release" version of mkFit @ PR #229). Assuming we want to drop phi1 and phi2, this would be as good a time as any, I imagine.

Note, we may need to adjust the axes in the multithreaded benchmark timing plots, because on lnx7188, the latest timings fall off the bottom of the plots!

kmcdermo:
@srlantz , in conjunction with #230 , I still have on my plate to revamp the scripts to drop phi1 and phi2 from our standard validation, but still have the option to run over them if needed.

srlantz:
One aspect of this issue is how to ensure scaling_governor is set correctly each time the benchmarks are run. The same type of care should be taken with the no_turbo setting. Alternatives for scaling_governor:

1. Have the script do "cpupower frequency-set -g performance" on each machine before it runs the benchmarks. However, this requires all users to have sudo privileges for this command.
2. Make this setting persistent on each machine with ""tuned-adm profile throughput-performance" per @dan131riley and have the script do "cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor" to verify that the setting has not changed.

The equivalent for no_turbo:

[...]

Issue #232, opened 2019-08-07
shift in multithreaded benchmarks due to new setting of CPU scaling governor on phi3


srlantz:
**Summary:**

On phi3, sometime between July 9 and 15, root ran the command "cpupower frequency-set -g performance", which affects CPU scaling behavior. The previous setting was "powersave". This change caused a shift in the multithreaded benchmark results on phi3; multithreaded timings are now significantly lower (better), and the biggest changes occur at high thread counts. 

**Action item:**

We need to decide whether to accept this as our new baseline, so that the new setting should be preserved going forward; and if so, we need to decide whether we should apply the same change to all test machines.

**Details:**

As noted in PR #229 on July 16, the customary benchmarks had showed an unexplained improvement in multithreaded performance on phi3. This applies only to phi3; performance on phi2 is unchanged. So far everyone has assumed that this is somehow due to the Intel compiler update that took place on July 15, which upgraded icc from 19.0.0 to 19.0.4 on phi3. However, @tresreid found that when he uses icc 19.0.0 to compile the HEAD of devel (currently the same as in PR #229 ), the timing performance does _not_ return to the previous baseline -

http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/July2019_AddValOpt/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png (PR #229 = HEAD, with 19.0.4 compiler)
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_performanceissues_08_08_19/benchmark_head_19-4/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png (HEAD with 19.0.4)
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_performanceissues_08_08_19/benchmark_head_19-0v2/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png (HEAD with 19.0.0)

In fact, on phi3, it seems it is currently impossible to reproduce _any_ of our older multithreaded benchmarks from on or before July 3 (when phi3 still had icc 19.0.0). Historically, on phi3, the multithreaded benchmark looks more like the following -

http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/208-Optional-HitSorting-Prefetch-Cleanup/Benchmarks/SKL-SP_CMSSW_TTbar_PU70_TH_time.png
[...]

Issue #230, opened 2019-07-19
Revamp of benchmark scripts


kmcdermo:
As discussed today, we decided to do the following from now for the default benchmarking and validation for both PRs and conferences:

- Drop phi1 and phi2 as test platforms (still available, but disabled by default)
- Drop SIMVAL and CMSSWVAL
- Add SIMVAL + MTV-Like, SIMVAL + MTV-Like w/ seed matching

I would like to raise the issue of #220 again. Do we want to process more events (or just fix the number of events?).  We have seen with more events processed / core, the thread utilization increases.  For PRs, so we don't take forever, we can still keep 20 evs / core.  For conference plots, we could increase this to 120 evs / core, and then do some of the manipulation that @makortel in his stress tests. 

The benchmarks are rather fast anyways, so bumping to 120 evts would not be the end of the world. The validation takes the longest since it is not really optimized.

kmcdermo:
PR #238 is merged, can make a new issue about moving to 120 events / thread.

PR #238, opened 2019-08-16
Benchmarking changes: drop phi1+2, change validation to MVT and check turbo and scaling settings

tresreid:
Changes to the benchmarking scripts.
1) phi1 and phi2 have been dropped from the default. The useLNX flag has been changed to useARCH flag. 
useARCH=0 (default) : phi3 only
useARCH=1: lnx4108 and lnx7188 only
useARCH=2: phi3 + lnx4108 and lnx7188
useARCH=3: phi1+phi2+phi3
useARCH=4: phi1+phi2+phi3 +lnx4108 +lnx7188

2) CMSSW val and SIM val have both been dropped. They have been replaced by MTV and MTV with required seeds

3) the turbo setting and scaling governor are printed out to the terminal when running. By default, the benchmarking script will crash if turbo is on or if the machine is not in performance mode. 

Benchmarks can be found here:
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_phidrop_08_16_19/benchmarks_Performance_phidrop_8_16_19v2/
http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_phidrop_08_16_19/benchmarks_Powersave_phidrop_8_16_19v2/


val_scripts/validation-cmssw-benchmarks.sh
```diff
@@ -55,16 +55,17 @@ bkfit="--backward-fit-pca"
 
 ## validation options: SIMVAL == sim tracks as reference, CMSSWVAL == cmssw tracks as reference
 SIMVAL="SIMVAL --sim-val ${siminfo} ${bkfit} ${style}"
-CMSSWVAL="CMSSWVAL --cmssw-val-fhit-bprm ${bkfit}"
-declare -a vals=(SIMVAL CMSSWVAL)
+SIMVAL_SEED="SIMVALSEED --sim-val ${siminfo} ${bkfit} --mtv-require-seeds"
```
kmcdermo, line 58: @tresreid , since we are dropping CMSSWVAL, might as well cleanup L59, L64

srlantz:
To enforce the agreed-upon hardware settings, namely no_turbo=1 and scaling_governor=performance, all 5 test machines are now equipped with the turbo-chmod service and have their tuned profile set to throughput-performance, as recommended by @dan131riley in the thread for issue #233. As a result, our preferred settings should now survive reboots. With these OS changes in place, everyone should now be able to switch to no_turbo=0 on a temporary basis (AND change it back!!!!), but only sudoers can mess with the scaling_governor.

web/collectBenchmarks.sh
```diff
@@ -79,7 +79,7 @@ done
 ######################################
 
 # Make SimTrack Validation directories
-simdir="SIMVAL"
+simdir="SIMVAL_MTV"
```
kmcdermo: @tresreid , since the contents of SIMVAL_MTV and SIMVAL_MTV_SEED will be the same, we can reduce the copy-paste, and just make a for loop over the two directories.

kmcdermo:
@tresreid , this looks really great -- thanks for doing this. This brings the following issues along

#233 - new baselines for scaling_governor (although I guess we will have to wait for phi1 to come back to life to close this for good) 
#232 - scaling_governor MT
#230 - drop phi1/phi2 + switch validation

I made a few comments, otherwise the changes look good to me. @srlantz , I think it makes sense if you review the code as well (and then hit the merge when you are satisfied and all requested changes are in + phi1 benchmarks).

kmcdermo:
@areinsvo , can you have a look at the MTV (+MTV-seed) validation plots as well? I will review the plots later today as well.

tresreid:
@kmcdermo I've made the changes you suggested.
I also plan to rerun the full benchmarks once all the machines are up and running so that the final benchmarks can all be in one place.

srlantz:
The script changes good to me. I'll let someone else (@areinsvo or @kmcdermo) do the merge after checking the physics performance plots, since those involve the newer metrics.

Comments on the benchmarks: as expected, the 3 SKL-SP machines show the same kind of dramatic improvement in multithreaded performance that we first saw (serendipitously) on phi3. KNL and SNB (in the plots that ave SNB data) don't show any real change. This is in line with expectations on KNL, which does almost no CPU frequency scaling. But I thought we might have seen some difference on SNB. (True, SKL-SP is much more aggressive about frequency scaling than SNB; SKL-SP can make separate adjustments on each core, and do so at a higher cadence.)

However, I think something might be wrong with the 8/16/19 SNB plots, because the same ones from 7/3/19 (when powersave was in effect) show significantly worse multithreaded performance on phi1, while the others are basically the same:

http://mireid.web.cern.ch/mireid/Benchmarks/benchmarks_7_3_19_saferdel/Benchmarks/

I wonder, was powersave mode ever really in effect on 8/16/19? It would be good to recheck when Tres presents the final-final benchmark plots for this PR.

kmcdermo:
> @kmcdermo I've made the changes you suggested.
> I also plan to rerun the full benchmarks once all the machines are up and running so that the final benchmarks can all be in one place.

Looks good to me, thanks for doing this. Once the final-final benchmarks are in, I will make sure to take a look (and also @areinsvo , if you could also look to confirm MTV looks good, just so we are at a nice clean baseline).

tresreid:
FYI: I've edited the original comment to put in the new benchmarks

areinsvo:
The MTV-like and MTV-require-seeds plots look consistent with the original set posted here: http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/July2019_AddValOpt

From my point of view this can be merged.

Issue #234, opened 2019-08-12
Possible bug in counting nFoundHits


areinsvo:
I've already mentioned this to a few people, but maybe it's good to keep track of officially too:
I ran the validation suite with mtv-require-seeds on the current head of devel and again after changing the validHitBonus from 10 to 500. The performance of STD is now fantastic (check out the plot vs number of layers), but it is totally different than the performance of CE. 

Bonus 10: http://areinsvo.web.cern.ch/areinsvo/OptV1/SIMVAL  
Bonus 50: http://areinsvo.web.cern.ch/areinsvo/OptV1_Bonus50/SIMVAL
Bonus 500: http://areinsvo.web.cern.ch/areinsvo/OptV1_Bonus500/SIMVAL 

There must be some difference in how the number of found hits is calculated between the two algorithms. Changing the hit penalty or the chi2 cut doesn’t cause any huge deviation between STD and CE. 

Looking at the nHits distributions seems to point to a bug in how NFoundHits is updated for STD:
http://areinsvo.web.cern.ch/areinsvo/OptV1_Bonus500/SIMVAL/nHits/
It’s worth pointing out that the variable being plotted here is NFoundHits_ taken from the Track object. There is no recounting of the hits associated to the track. So it’s possible that the track has a reasonable number of hits but NFoundHits_ isn’t being updated correctly.

kmcdermo:
Wanted to link here @areinsvo slides from today's call, highlighting this issue on slide 4: https://indico.cern.ch/event/827099/contributions/3460338/attachments/1894190/3124594/OptimizeEfficiency_Aug2019.pdf

For now, we chose not to have crazy hit bonus values (which, as Allie demonstrated, we want to be lower anyways) and just print a warning in case it does happen, and then wait until Track Kaboom goes into place to resolve this for good.

areinsvo:
This issue was kept open to remind us that the candidate scores can overflow if you try nonsensical values (like 5000) for the found hit bonus. This is resolved after [Track Kaboom in PR243](https://github.com/trackreco/mkFit/pull/243) because score is now stored as a float.

Issue #231, opened 2019-08-02
Recover efficiency for long tracks


areinsvo:
As announced previously and discussed in the meeting today, the fixes in hit counting (see [PR here](https://github.com/trackreco/mkFit/pull/214) ) that were merged in April changed our physics performance. We do ***worse*** for long tracks than we did before. This was shown in [this presentation](https://indico.cern.ch/event/783876/contributions/3261825/attachments/1820631/2977630/HitCounting_March29.pdf), for example.

Compare the [results from before the changes to hit counting](http://areinsvo.web.cern.ch/areinsvo/MkFit/DuplicateRemoval/forPR/withDuplicateRemoval/SIMVAL/SIMVAL2/) with the [results after the changes](http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/July2019_AddValOpt/SIMVAL/). Both results are with our standard stand-alone validation definitions, which haven't changed. 

We should make an effort to recover this efficiency.

areinsvo:
As demonstrated in [PR239](https://github.com/trackreco/mkFit/pull/239) and [PR241](https://github.com/trackreco/mkFit/pull/241), this efficiency loss was recovered by retuning the parameters used for the score calculation. While a final tuning will be needed at some point, I think it is safe to close this issue.

Issue #195, opened 2018-12-10
Recovering efficiency for short tracks


cerati:
This issue is to track ideas to try to recover the efficiency for short tracks. Note that these are not monitored in the standard validation, see issue #193 to find the instructions to lower our selection.

[...]

PR #239, opened 2019-08-20
Update score calc, add warning for score overflow

areinsvo:
This PR addresses things that were discussed in issues [#195](https://github.com/trackreco/mkFit/issues/195), [#231](https://github.com/trackreco/mkFit/issues/231), and [#234](https://github.com/trackreco/mkFit/issues/234).

By simplifying the score calculation and reducing the hit bonus from 10 to 5, we see a large improvement in efficiency and a slight improvement in timing using our new MTV-require-seeds validation. This is a preliminary retuning, but I propose to use it as the new baseline for further optimizations.

The standalone results can be found [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/OptimizeAug2019/), for both the previous validation definitions and the mtv-require-seeds validation. The results when run within CMSSW_10_4 can be found [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization).

See also the slides presented in one of our mkFit meetings [here](https://indico.cern.ch/event/827099/contributions/3460338/attachments/1894190/3124594/OptimizeEfficiency_Aug2019.pdf).

### DIFF
--- a/Config.h
+++ b/Config.h
@@ -359,7 +359,7 @@ namespace Config
   constexpr float track_ptlow = 0.9;
 
   // sorting config (bonus,penalty)
-  constexpr float validHitBonus_ = 10.0;//2.5*4.0; // 4x cmssw bonus (after fix for counts of # missing hits)
+  constexpr float validHitBonus_ = 5.0;//2.5*4.0; // 4x cmssw bonus (after fix for counts of # missing hits)
   constexpr float missingHitPenalty_ = 5.0;//20.0/4.0; // 0.25x cmssw penalty (after fix for counts of # missing hits)  
   constexpr float maxChi2ForRanking_ = 819.2f; // (=0.5f*0.1f*pow(2,14);)
 
--- a/Config.h
+++ b/Config.h
@@ -359,8 +359,8 @@ namespace Config
   constexpr float track_ptlow = 0.9;
 
   // sorting config (bonus,penalty)
-  constexpr float validHitBonus_ = 5.0;//2.5*4.0; // 4x cmssw bonus (after fix for counts of # missing hits)
-  constexpr float missingHitPenalty_ = 5.0;//20.0/4.0; // 0.25x cmssw penalty (after fix for counts of # missing hits)  
+  constexpr float validHitBonus_ = 5.0;//cmssw bonus = 2.5
+  constexpr float missingHitPenalty_ = 5.0;//cmssw penalty = 20
   constexpr float maxChi2ForRanking_ = 819.2f; // (=0.5f*0.1f*pow(2,14);)
 
   // Threading
--- a/Track.h
+++ b/Track.h
@@ -547,6 +547,9 @@ inline int getScoreCalc(const unsigned int seedtype,
   //if(chi2>Config::maxChi2ForRanking_) chi2=Config::maxChi2ForRanking_;
   int score = 0;
   float score_ = 0.f;
+  ////// V2 of candidate score (simplified score, after fix for counts of # missing hits):
+  score_ = Config::validHitBonus_*nfoundhits - Config::missingHitPenalty_*nmisshits - chi2;
+  if(pt<0.9f && seedtype==2) score_ -= 0.5f*(Config::validHitBonus_)*nfoundhits;
   /*
   ////// V0 of candidate score (before fix for counts of # missing hits):
   // For high pT central tracks: double valid hit bonus
@@ -570,6 +573,7 @@ inline int getScoreCalc(const unsigned int seedtype,
   */
   ////// V1 of candidate score (after fix for counts of # missing hits):
   // For high pT central tracks: 4x valid hit bonus and 0.25x missing hit penalty
+  /*
   if(seedtype==1){
     score_ = (Config::validHitBonus_*4.0f)*nfoundhits - Config::missingHitPenalty_*nmisshits*0.25f - chi2;
     if(pt<0.9f) score_ -= 0.25f*(Config::validHitBonus_)*nfoundhits;
@@ -587,6 +591,8 @@ inline int getScoreCalc(const unsigned int seedtype,
     if(pt<0.9f) score_ -= 0.2f*(Config::validHitBonus_)*nfoundhits;
     if(nfoundhits>8) score_ += (Config::validHitBonus_)*nfoundhits;
   }
+  */
+  if(score > 16000 && !Config::silent) std::cout << "WARNING!! Might be facing score overflow, which leads to discrepancies between STD and CE!" << std::endl;
   score = (int)(floor(10.f * score_ + 0.5));
   return score;
 }
--- a/Track.h
+++ b/Track.h
@@ -592,7 +592,6 @@ inline int getScoreCalc(const unsigned int seedtype,
     if(nfoundhits>8) score_ += (Config::validHitBonus_)*nfoundhits;
   }
   */
-  if(score > 16000 && !Config::silent) std::cout << "WARNING!! Might be facing score overflow, which leads to discrepancies between STD and CE!" << std::endl;
   score = (int)(floor(10.f * score_ + 0.5));
   return score;
 }
--- a/plotting/PlotValidation.cpp
+++ b/plotting/PlotValidation.cpp
@@ -1034,7 +1034,7 @@ void PlotValidation::SetupBins()
   PlotValidation::SetupFixedBins(110,0,1.1,fFracHitsBins);
 
   // track score bins
-  PlotValidation::SetupFixedBins(1699,-16990,16990,fScoreBins);
+  PlotValidation::SetupFixedBins(50,-500,5000,fScoreBins);
 
   // dNhits
   PlotValidation::SetupFixedBins(40,-20,20,fDNHitsBins);


kmcdermo:
Thanks for putting this up! I will take a look at the code shortly. 
Two quick things: 
1. @areinsvo , Can you rebase to the HEAD of devel and then rerun the benchmarks (and attach them here)? This way we can compare directly the plots from PR #238 .
2. I will add some slides later today that tries to tease out what exactly happened with the mkFit standalone validation.

makortel:
I took a look of the MTV plots in http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_initialStep.html
The legend for the plots is
* blue for CMSSW
* red for mkFit before this PR
* black for mkFit with this PR

Efficiency looks good across the board (much better than before, also a bit better than CMSSW)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/effandfakePtEtaPhi.pdf

especially for short tracks (< 12 hits / 8 layers of SimTrack)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/effandfakeHitsLayers.pdf

The fake rate is now very close to CMSSW, but duplicate rate about doubles
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/dupandfakePtEtaPhi.pdf

There are a couple of corners of phase space where the fakes increase (this time showing distributions of tracks instead of rates). One place is the "track reference point radius wrt. beamspot" in the region of 0.8-2 mm (IIRC the reference point is essentially the point of closes approach to the beam line)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/distPosDeltaR.pdf

Another place is the minimum DeltaR between tracks, which is also interesting for duplicates. The plot below shows all matched tracks on the left, fake tracks in the middle, and duplicate tracks on the right. Essentially the added duplicate tracks become visible for DeltaR < 0.02, while for fakes (in the black curve) there is a clear kink in the distribution around DeltaR 0.04
image

The tracks contain significantly less hits now wrt. both earlier and CMSSW (these plots are for sim-matched tracks)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/hitsAndPt.pdf
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/hitsLayers.pdf

The residuals of phi, cot(theta), dxy, and dz are essentially the same between all three cases, but pT shows some difference between mkFit and CMSSW (mkFit now and before are essentially the same). MkFit is closer to zero for endcaps (|eta| > 1.4), and CMSSW is closer to zero for barrel (|eta| < 0.8)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/residualEta.pdf
The residuals have also pT dependence, mkFit staying closer to zero for much lower pT values (0.3 GeV vs. 1 GeV or so)
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/residualPt.pdf

Resolutions show differences in all parameters, with mkFit being now worse than before. Worst case is the resolution of pT
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/resolutionsEta.pdf
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/resolutionsPt.pdf

Pulls of phi, theta, dxy, and dz are tiny bit wider for mkFit vs. CMSSW (mkFit now and before being essentially the same), for pT and q/p there are clear differences with mkFit having narrower tails now than before, but being still wider than CMSSW
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/pulls.pdf

The (normalized) chi2 had a mean larger than CMSSW before, but is now smaller. Likewise, the chi2 probability distribution was flatter than CMSSW before, but is now less flat.
image
http://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/Aug2019_Optimization/plots_compareMkFit/plots_building_initialStep/tuning.pdf

kmcdermo:
@areinsvo , can you also please change the width of the score plots, now that they are somewhat sensible as seen here: http://areinsvo.web.cern.ch/areinsvo/MkFit/OptimizeAug2019/SIMVAL_MTV-require-seeds/score/SKL-SP_CMSSW_TTbar_PU70_allreco_score_build_pt0p0_SIMVAL.png

The relevant lines are here: https://github.com/trackreco/mkFit/blob/e6feec69756a2039bbbe2662541b9e4714fc3dd4/plotting/PlotValidation.cpp#L1036-L1037

I think a more reasonable range is now 0-5000, maybe with bin width of 50?

Config.h
```diff
@@ -359,7 +359,7 @@ namespace Config
   constexpr float track_ptlow = 0.9;
 
   // sorting config (bonus,penalty)
-  constexpr float validHitBonus_ = 10.0;//2.5*4.0; // 4x cmssw bonus (after fix for counts of # missing hits)
+  constexpr float validHitBonus_ = 5.0;//2.5*4.0; // 4x cmssw bonus (after fix for counts of # missing hits)
```
slava77: the comment is inconsistent now

Track.h
```diff
@@ -587,6 +591,8 @@ inline int getScoreCalc(const unsigned int seedtype,
     if(pt<0.9f) score_ -= 0.2f*(Config::validHitBonus_)*nfoundhits;
     if(nfoundhits>8) score_ += (Config::validHitBonus_)*nfoundhits;
   }
+  */
+  if(score > 16000 && !Config::silent) std::cout << "WARNING!! Might be facing score overflow, which leads to discrepancies between STD and CE!" << std::endl;
```
slava77: this will get messed up in multithreaded runs.
also, was this function in a vectorized part?
the cout is another badness.

Track.h
```diff
@@ -570,6 +573,7 @@ inline int getScoreCalc(const unsigned int seedtype,
   */
   ////// V1 of candidate score (after fix for counts of # missing hits):
   // For high pT central tracks: 4x valid hit bonus and 0.25x missing hit penalty
+  /*
```
slava77, line 576: instead of commenting out a better practice would be to allow to choose with a switch
or simply remove if versioning is not to be really supported

Track.h
areinsvo: @slava77 do you have a suggestion for a better way to handle this?

slava77: we have some print and dprint options.
@dan131riley @osschar I think that you know better than me.

areinsvo:
The new benchmarks after the rebase can be found here:
http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR239

Track.h
areinsvo: When doing the tuning, it was useful that @mmasciov left V0 in the code but commented out, so that I could easily see what changed from V0 to V1. But that could also be checked in the github history, so I'm fine with removing the V0 and V1 comment blocks if @mmasciov agrees.

Track.h
dan131riley: The Track class is an awkward place to report an error.  I guess I'd be in favor of adding a `static std::atomic<int>` error counter to the Track class.  Make the message a debug one with `dprint` and report any non-zero error counts at the end with the summary statistics.

kmcdermo:
EDIT: New slides attached to fix up some wording with nHits <--> nLayers
EDIT 2: Included a few new plots in backup for pT > 0.9 that mirror plots in main talk (overrode link to slides). Full dump in backup is only for pT > 0.0.

Hi all,

After consulting with Giuseppe and Allie, we have put together some [slides](https://github.com/trackreco/mkFit/files/3535427/before_after_simval_bonus.pdf) on the discrepancy between MTV and mkFit validation from updating the track score calculation.

The main takeaway is that the min number of reco layers in the mkFit validation was fooling us in the following ways: 
1. Making us blind to short reco track eff / FR
2. When optimizing our score, tracks that aggressively added hits were prioritized over those that did not

By simply relaxing the number of required reco layers, we see that we can "recover" some efficiency with the new score, as short reco tracks can now be matched to long sim tracks. This puts us more in line with what we have seen before and the new MTV improvements.

Please take some time to digest them... I imagine this will be a topic of some discussion on Friday. Post any comments you have to the GH chat so it can all be in one place.

kmcdermo:
As @makortel out, on slide 4 "nHits" for sim track should be "nLayers", and "nLayers" for reco tracks should be "nHits".

Config.h
areinsvo: I changed the comment to state what the cmssw values are, instead of having the comment be relative to the mkFit values.

Track.h
areinsvo: In the interest of getting this merged ASAP, I removed the cout statement. This should be addressed in a different PR before we close issue [#234](https://github.com/trackreco/mkFit/issues/234)

areinsvo:
From my point of view, this is ready to merge.

kmcdermo:
Hi all, I propose to merge this PR within a few hours, barring any objections.

There are still some open points that can go into separate issues  / PRs, namely:
- Provide warnings for crazy track score by printing sum at end using an std::atomic counter 
- Assess other score values to increase nHits / track
- Further cleanups of V0/V1 if needed 

As a reminder, this is simply a baseline for which we can improve.

osschar:
The plan is to use float for the score when we change data format. So this is really a transient issue ... we put the score in like this to avoid changing the Track class :)

kmcdermo:
> The plan is to use float for the score when we change data format. So this is really a transient issue ... we put the score in like this to avoid changing the Track class :)

Right ... I meant to mention that, so thank you for reminding me :)

PR #241, opened 2019-08-29
Change found hit bonus to 7.5

areinsvo:
Changing the hit bonus to 7.5 maintains a track building efficiency comparable to CMSSW but improves the number of strip layers in the built tracks (compared to the hit bonus of 5). Obviously this still doesn't represent a fine tuning, but it is a better set of comparison plots to use for the POG update than my initial guess of 5. Mario said he might have time to work on a further optimization in the next week.

Standalone validation plots: http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/standalone
MTV plots (clean version): http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/plots_NewScore_Bonus7p5 
MTV plots (messy version, with old mkFit score, new mkFit score with bonus of 5, and new mkFit score with bonus of 7.5): http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/plots_compareScores/
A few notable plots include the efficiency comparison [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/plots_compareScores/plots_building_initialStep/effandfakePtEtaPhi.pdf) and the comparison of the number of layers in the built tracks [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/plots_compareScores/plots_building_initialStep/hitsLayers.pdf).

### DIFF
--- a/Config.h
+++ b/Config.h
@@ -359,7 +359,7 @@ namespace Config
   constexpr float track_ptlow = 0.9;
 
   // sorting config (bonus,penalty)
-  constexpr float validHitBonus_ = 5.0;//cmssw bonus = 2.5
+  constexpr float validHitBonus_ = 7.5;//cmssw bonus = 2.5
   constexpr float missingHitPenalty_ = 5.0;//cmssw penalty = 20
   constexpr float maxChi2ForRanking_ = 819.2f; // (=0.5f*0.1f*pow(2,14);)
 


kmcdermo:
A bonus of 7.5 definitely puts us much closer to CMSSW in nLayers, completely within uncertainties for the barrel, and now only down on average 2 layers in the endcaps (compared to missing 2 layers in the barrel and 4 in the endcaps for the hit bonus of 5).

This is definitely a better point to show to CMS during the Tracking POG meetings / CMS week.

slava77:
@areinsvo 
did you try a larger value (I'm most interested in the old 10)?
I'm curious how close are we from what seems like a cliff in efficiency for tracks with pt 1.-2 GeV
http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/plots_compareScores/plots_building_initialStep/effandfakePtEtaPhi.pdf
Screen Shot 2019-08-29 at 5 29 04 AM
assuming that most seeds for tracks pt in 1-2GeV range are above 0.9 GeV, then the only difference with red is the valid hit bonus (5 for black, 7.5 for orange, and 10 for red). It looks like stepping from 5 to 7.5 has a factor of about 5 smaller effect on efficiency than stepping from 7.5 to 10.
... OTOH, it could be that there is something peculiar going on with seed pt below 0.9 GeV in the endcaps for the tracks with true pt 1.-2 GeV

cerati:
@slava77 just to point out that there is indeed something "special" with a valid bonus of 10, since the chi2cut is 15 and the invalid penalty is 5. So it's not unlikely that there is a "cliff in efficiency" nearby. See Allie's slides for a more complete discussion (https://indico.cern.ch/event/827099/contributions/3460338/attachments/1894190/3124594/OptimizeEfficiency_Aug2019.pdf)

slava77:
> @slava77 just to point out that there is indeed something "special" with a valid bonus of 10, since the chi2cut is 15 and the invalid penalty is 5. So it's not unlikely that there is a "cliff in efficiency" nearby. See Allie's slides for a more complete discussion (https://indico.cern.ch/event/827099/contributions/3460338/attachments/1894190/3124594/OptimizeEfficiency_Aug2019.pdf)

this implies a large phase space of candidates in the range between "validBonus + penalty" and the chi2 cut off of 15. Currently it's 12.5 vs 15. Naively, this phase space should be small though.

cerati:
I do not claim to understand all details, but a difference of 2.5 is not so small in the phase space of parameter values of <15.

areinsvo:
For the sake of documentation: @slava77 the comparison between red and black/orange is not completely straightforward, because the hit bonus changed but we also changed from "V1" to "V2" of the candidate score. V1 had additional categories that tweaked the bonus or penalties, so in reality the bonus was at least 17 in each category.

slava77:
> For the sake of documentation: @slava77 the comparison between red and black/orange is not completely straightforward, because the hit bonus changed but we also changed from "V1" to "V2" of the candidate score. V1 had additional categories that tweaked the bonus or penalties, so in reality the bonus was at least 17 in each category.

ah, good point. 
I forgot and then while checking the diff did not notice ```/* ```.

kmcdermo:
Have we decided we want to make this our baseline, given that we are using this for the Sep. 9 TRK POG meeting? Or are we waiting for further optimization?

mmasciov:
I believe this is a better baseline, and we can take it as a landmark for the presentation.

M.

Issue #236, opened 2019-08-13
Integration to CMSSW master


makortel:
This issue is to track progress towards integrating mkFit to CMSSW master:

Overall (to get mkFit running within offline reco)
- [x] merge #235
- [x] maybe include #237 as well?
- [x] tag a new release
- [x] create mkFit external to [cmsdist](https://github.com/cms-sw/cmsdist), merged as https://github.com/cms-sw/cmsdist/pull/5150
   - [x] finalize mkFit external, merged as https://github.com/cms-sw/cmsdist/pull/5169
- [x] tag a new release following the merge of #241 
- [x] update mkFit external in cmsdist, merged as https://github.com/cms-sw/cmsdist/pull/5213
- [x] integrate `RecoTracker/MkFit` package from [makortel:mkfit_1040p1](https://github.com/makortel/cmssw/tree/mkfit_110x_integ) to CMSSW master, merged as https://github.com/cms-sw/cmssw/pull/27895
- [x] update mkFit with interface towards CMSSW to enable duplicate removal, done in #244
- [x] enable duplicate removal in `MkFitProducer`
- [x] Deal with non-x86 architectures #240
- [x] Deal with OpenMP runtime lib https://github.com/trackreco/mkFit/issues/257
- [x] Deal with SMatrix https://github.com/trackreco/mkFit/issues/258 (done in #264)
- [x] Fix backward fit for triplet seeds (#265)
- [ ] update mkFit external

Getting mkFit to run within HLT reconstruction some additional changes may be wanted
- [x] either re-open and merge https://github.com/cms-sw/cmssw/pull/26322, or remove or allow chancing the minimum hit cut for seeds in the seed cleaning, merged as https://github.com/cms-sw/cmssw/pull/28473 with a configuration flag https://github.com/trackreco/mkFit/blob/9afc251699bc3f07d4a02320ce229c25a8cffca9/Event.cc#L855
   * the seeds from quadruplet pixel tracks (as in HLT iter0 at the moment) get their last hit dropped, so the seed cleaning will drop all seeds in that case
   * of course all this will change for run3
- [x] make PR for https://github.com/makortel/cmssw/commit/ddd42d7c74ecabd9d4bf3b649bc40881ed76e8a0 (Add a flag to disable strip rechit matching), merged as https://github.com/cms-sw/cmssw/pull/28471
   * will become close-to-unnecessary once we get the GPU version of strip local reco
- [x] add customize function to use mkFit in HLT iter0 pattern recognition

[...]

PR #242, opened 2019-09-10
Add support pixel triplet seeds

makortel:
This PR is to add support for pixel triplet seeds in order to work around a "feature" of CMSSW EDProducer that converts pixel reco::Tracks to TrajectorySeeds (used in the HLT menu today) that ignores the fourth hit of the seed (in the context of #236).

First step is to remove the "minimum number of hits in a seed" cut from the seed cleaner, but that turned out to be insufficient. The story continues in the comments.

### DIFF
--- a/mkFit/MkBuilder.cc
+++ b/mkFit/MkBuilder.cc
@@ -189,25 +189,27 @@ MkBuilder::MkBuilder() :
 
   { SteeringParams &sp = m_steering_params[TrackerInfo::Reg_Endcap_Neg];
     sp.reserve_plan(3 + 3 + 6 + 18);
-    sp.fill_plan(0, 2, false, true);
-    sp.append_plan(45, true);
+    sp.fill_plan(0, 1, false, true); // bk-fit only
+    sp.append_plan( 2, true);        // pick-up only
+    sp.append_plan(45, false);
     sp.append_plan(46, false);
     sp.append_plan(47, false);
-    sp.fill_plan(48, 53); // TID, 6 layers
+    sp.fill_plan(48, 53); // TID,  6 layers
     sp.fill_plan(54, 71); // TEC, 18 layers
     sp.finalize_plan();
   }
 
   { SteeringParams &sp = m_steering_params[TrackerInfo::Reg_Transition_Neg];
     sp.reserve_plan(3 + 4 + 6 + 6 + 8 + 18);
-    sp.fill_plan(0, 2, false, true);
-    sp.append_plan( 3, true);
-    sp.append_plan(45, true);
+    sp.fill_plan(0, 1, false, true); // bk-fit only
+    sp.append_plan( 2, true);
+    sp.append_plan( 3, false);
+    sp.append_plan(45, false);
     sp.append_plan(46, false);
     sp.append_plan(47, false);
-    sp.fill_plan( 4,  9); // TIB, 6 layers
-    sp.fill_plan(48, 53); // TID, 6 layers
-    sp.fill_plan(10, 17); // TOB, 8 layers
+    sp.fill_plan( 4,  9); // TIB,  6 layers
+    sp.fill_plan(48, 53); // TID,  6 layers
+    sp.fill_plan(10, 17); // TOB,  8 layers
     sp.fill_plan(54, 71); // TEC, 18 layers
     sp.finalize_plan();
   }
@@ -215,34 +217,37 @@ MkBuilder::MkBuilder() :
 
   { SteeringParams &sp = m_steering_params[TrackerInfo::Reg_Barrel];
     sp.reserve_plan(3 + 1 + 6 + 8);
-    sp.fill_plan(0, 2, false, true);
-    sp.append_plan(3, true); // pickup-only
-    sp.fill_plan( 4,  9);    // TIB, 6 layers
-    sp.fill_plan(10, 17);    // TOB, 8 layers
+    sp.fill_plan(0, 1, false, true); // bk-fit only
+    sp.append_plan( 2, true);        // pickup-only
+    sp.append_plan( 3, false);
+    sp.fill_plan( 4,  9); // TIB, 6 layers
+    sp.fill_plan(10, 17); // TOB, 8 layers
     sp.finalize_plan();
   }
 
   { SteeringParams &sp = m_steering_params[TrackerInfo::Reg_Transition_Pos];
     sp.reserve_plan(3 + 4 + 6 + 6 + 8 + 18);
-    sp.fill_plan(0, 2, false, true);
-    sp.append_plan( 3, true);
-    sp.append_plan(18, true);
+    sp.fill_plan(0, 1, false, true); // bk-fit only
+    sp.append_plan( 2, true);        // pickup-only
+    sp.append_plan( 3, false);
+    sp.append_plan(18, false);
     sp.append_plan(19, false);
     sp.append_plan(20, false);
-    sp.fill_plan( 4,  9); // TIB, 6 layers
-    sp.fill_plan(21, 26); // TID, 6 layers
-    sp.fill_plan(10, 17); // TOB, 8 layers
+    sp.fill_plan( 4,  9); // TIB,  6 layers
+    sp.fill_plan(21, 26); // TID,  6 layers
+    sp.fill_plan(10, 17); // TOB,  8 layers
     sp.fill_plan(27, 44); // TEC, 18 layers
     sp.finalize_plan();
   }
 
   { SteeringParams &sp = m_steering_params[TrackerInfo::Reg_Endcap_Pos];
     sp.reserve_plan(3 + 3 + 6 + 18);
-    sp.fill_plan(0, 2, false, true);
-    sp.append_plan(18, true);
+    sp.fill_plan(0, 1, false, true); // bk-fit only
+    sp.append_plan( 2, true);        // pickup-only
+    sp.append_plan(18, false);
     sp.append_plan(19, false);
     sp.append_plan(20, false);
-    sp.fill_plan(21, 26); // TID, 6 layers
+    sp.fill_plan(21, 26); // TID,  6 layers
     sp.fill_plan(27, 44); // TEC, 18 layers
     sp.finalize_plan();
   }
@@ -782,10 +787,7 @@ inline void MkBuilder::fit_one_seed_set(TrackVec& seedtracks, int itrack, int en
 
   if (Config::cf_seeding) mkfttr->ConformalFitTracks(false, itrack, end);
 
-  if (Config::seedInput != cmsswSeeds)
-  {
-    mkfttr->FitTracksSteered(is_brl, end - itrack, m_event, Config::seed_fit_pflags);
-  }
+  mkfttr->FitTracksSteered(is_brl, end - itrack, m_event, Config::seed_fit_pflags);
 
   mkfttr->OutputFittedTracksAndHitIdx(m_event->seedTracks_, itrack, end, false);
 }
@@ -1545,7 +1547,12 @@ void MkBuilder::PrepareSeeds()
   //0 = not set; 1 = high pT central seeds; 2 = low pT endcap seeds; 3 = all other seeds
   assign_seedtype_forranking();
 
-  fit_seeds();
+  // Do not refit cmssw seeds (this if was nested in fit_one_seed_set() until now).
+  // Eventually we can add force-refit option.
+  if (Config::seedInput != cmsswSeeds)
+  {
+    fit_seeds();
+  }
 }
 
 //------------------------------------------------------------------------------
--- a/Config.h
+++ b/Config.h
@@ -331,7 +331,6 @@ namespace Config
   extern float maxdRSquared;
 
   // config on seed cleaning
-  constexpr int minNHits_seedclean = 4;
   constexpr float track1GeVradius = 87.6; // = 1/(c*B)
   constexpr float c_etamax_brl = 0.9;
   constexpr float c_dpt_brl_0 = 0.025;
--- a/Event.cc
+++ b/Event.cc
@@ -808,7 +808,6 @@ void Event::print_tracks(const TrackVec& tracks, bool print_hits) const
 int Event::clean_cms_seedtracks()
 {
 
-  const int minNHits     = Config::minNHits_seedclean;
   const float etamax_brl = Config::c_etamax_brl;
   const float dpt_brl_0  = Config::c_dpt_brl_0;
   const float dpt_ec_0   = Config::c_dpt_ec_0;
@@ -871,7 +870,6 @@ int Event::clean_cms_seedtracks()
   for(int ts=0; ts<ns; ts++){
 
     if (not writetrack[ts]) continue;//FIXME: this speed up prevents transitive masking; check build cost!
-    if (nHits[ts] < minNHits) continue;
 
     const float oldPhi1 = oldPhi[ts];
     const float pos2_first = pos2[ts];
@@ -881,8 +879,6 @@ int Event::clean_cms_seedtracks()
 
     //#pragma simd /* Vectorization via simd had issues with icc */
     for (int tss= ts+1; tss<ns; tss++){
-
-      if (nHits[tss] < minNHits) continue;
 
       const float Pt2 = pt[tss];
 
--- a/README.md
+++ b/README.md
@@ -559,13 +559,13 @@ named as initialStep)
 ```bash
 # in CMSSW_10_4_0_patch1/src
 
-# sample = 10mu, ttbarnopu, ttbar_pu50, ttbar_pu70
+# sample = 10mu, ttbarnopu, ttbarpu50, ttbarpu70
 # mkfit = 1, 0
 # timing = 0, 1
 # (maxEvents = 0, <N>, -1)
 # nthreads = 1, <N>
 # nstreams = 0, <N>
-cmsRun RecoTracker/MkFit/test/reco_cfg.py sample=ttbar_pu70 timing=1
+cmsRun RecoTracker/MkFit/test/reco_cfg.py sample=ttbarpu50 timing=1
 ```
 * The default values for the command line parameters are the first ones.
 * `mkfit=1` runs MkFit, `0` runs CMSSW tracking
@@ -608,9 +608,20 @@ HLT reconstruction
 ```bash
 # in CMSSW_10_4_0_patch1/src
 
-cmsRun RecoTracker/MkFit/test/hlt_cfg.py sample=ttbar_pu70 timing=1
+# in addition to the offline tracking options
+# hltOnDemand = 0, 1
+# hltIncludeFourthHit = 0, 1
+cmsRun RecoTracker/MkFit/test/hlt_cfg.py sample=ttbarpu50 timing=1
 ```
-* Options and behavior is the same as for the offline reconstruction above
+* The default values for the command line parameters are the first ones.
+* For options that are same as in offline tracking, see above
+* Setting `hltOnDemand=1` makes the strip local reconstruction to be
+  run in the "on-demand" mode (which is the default in real HLT but
+  not here). Note that `hltOnDemand=1` works only with `mkfit=0`.
+* Setting `hltIncludeFourthHit=1` changes the (HLT-default) behavior
+  of the EDProducer that converts pixel tracks to `TrajectorySeed`
+  objects to include also the fourth, outermost hit of the pixel track
+  in the seed.
 
 DQM harvesting (unless running timing)
 ```bash


makortel:
The benchmarks are here
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242/

makortel:
This PR (as of now) works fine in the standalone program. When ran from CMSSW with triplet seeds (made from quadruplet pixel tracks, as they are provided within the HLT menu), I get a segfault in
```
#4  <signal handler called>
#5  Matriplex::MatriplexSym<float, 3, 16>::CopyIn (this=0x7fe97bf1e440, n=<optimized out>, arr=0xc) at ../Matriplex/MatriplexSym.h:98
#6  mkfit::MkFitter::InputTracksAndHits (this=0x7fe97bf1d1c0, tracks=..., layerHits=..., beg=0, end=2079450176) at MkFitter.cc:155
```
i.e.
https://github.com/trackreco/mkFit/blob/d9dfbda14e228b3d2dc229e9cd3339bb1f529139/mkFit/MkFitter.cc#L149-L155
The `Nhits` is set in
https://github.com/trackreco/mkFit/blob/d9dfbda14e228b3d2dc229e9cd3339bb1f529139/mkFit/MkFitter.h#L60
https://github.com/trackreco/mkFit/blob/d9dfbda14e228b3d2dc229e9cd3339bb1f529139/mkFit/MkBuilder.cc#L775-L780
where `Config::nlayers_per_seed` ultimately comes from
https://github.com/trackreco/mkFit/blob/d9dfbda14e228b3d2dc229e9cd3339bb1f529139/Geoms/CMS-2017.cc#L19

Do I understand correctly that we are setting globally the exact number of hits per seed instead of a maximum?

Any suggestions how to proceed?

@osschar @slava77

slava77:
> Any suggestions how to proceed?

an "easy" option now would be to get CMSSW (configurably) do what we want.
However, things will probably go bad again after we start getting a merged seed collection from patatrackers.

kmcdermo:
@makortel , this is set during the fit over seeds before processing, which uses the MkFitter routines, which expect an exact number of layers (as the fitting goes layer by layer). 

@osschar , correct me if I am wrong -- did you mention on that with Track kaboom this should no longer really problem?

osschar:
Why are we fitting seeds here -- this is where the crash happens? For patatrack seeds in our bin files we use the existing track params as input ... we do not refit them. The seed fit is only used for sim track seeds ... and there we do expect them to have 4 layers, indeed, as we construct the seeds by picking the first four hits from sim-track that lay on unique layers, dropping overlaps.

Saying this, there are apparently two things for me to look at:
1. Fix the seed fit to also handle 3-hit input.
2. Investigate how LayerPlan has to be modified to properly support 3-hit tracks, especially those that are missing the hit on the last layer ... as steering params are configured right now to ignore the last layer for hit-pickup / finding.

@kmcdermo no ... this has nothing to do with Track Kaboom

kmcdermo:
I apologize for leading people astray on this. The last step of MkBuilder::PrepareSeeds() is MkBuilder::fit_seeds() (which is always called), and inside that is MkBuilder::fit_one_seed_set()... 

But as already mentioned, inside that last function there is a check if we are using CMSSW seeds, in which case if we are then the fitting is skipped and the seeds are imported directly into MkFitters. Sorry about that. 

The crash comes from the MkFitter::InputTracksAndHits() in MkBuilder::find_one_seed_set().

osschar:
I disabled fitting for cmssw seeds at an earlier stage (so we never get to the piece of code where we had the crash) and fixed steering-params so that seeds are picked up also on earlier layers (this now expects the seeds to end on layer 2 (or later), i.e., if there is a doublet seed with hits on layers 0 and 1 it will not be picked up).

https://github.com/trackreco/mkFit/tree/store-hits-per-seed

@makortel can you give it a try together with your changes here and with your triplet dataset?

osschar:
Grr, I posted a wrong branch:
https://github.com/trackreco/mkFit/tree/seed-fit-fix

makortel:
> https://github.com/trackreco/mkFit/tree/seed-fit-fix

@osschar Thanks. I (in my private repo) cherry-picked that on top of this branch (and rebased on top of `devel`) and tried again. Now the HLT job fails with
```
An exception of category 'BadAlloc' occurred while
   [0] Processing  Event run: 1 lumi: 1 event: 17 stream: 0
   [1] Running path 'MC_ReducedIterativeTracking_v12'
   [2] Calling method for module MkFitProducer/'hltIter0PFlowCkfTrackCandidatesMkFit'
Exception Message:
A std::bad_alloc exception was thrown.
The job has probably exhausted the virtual memory available to the process.
```
But that seems to be caused by #243 (and happens in the offline reconstruction configuration as well). Or, at least if I take my branch of #244, the offline reco works, if I take `devel`, I get the `std::bad_alloc` exception.

makortel:
> But that seems to be caused by #243

Probably because I did not do the necessary updates on the CMSSW side yet...

osschar:
Do you need me to do / check something? :)

makortel:
So let me try again
> https://github.com/trackreco/mkFit/tree/seed-fit-fix

I (in my private repo) cherry-picked that on top of this branch (and rebased on top of #244) and tried again. At least technically the HLT workflow works now, need to check the physics performance next.

makortel:
> Do you need me to do / check something? :)

Let me remind myself first has to be done from #243 first :)

makortel:
@osschar I finally got to complete the test, and https://github.com/trackreco/mkFit/tree/seed-fit-fix combined with this PR seems to work fine with pixel triplet seeds. How sould we proceed? (cherry-pick your commit here, two separate PRs, something else?)

IHateLinus:
Hi Matti.
This is really good. Do we have a sense of (physics) performance of these?

makortel:
Let me quote now only efficiency vs. pT, I'll rerun with the full 10k event to have more precise picture
image
Overall looks pretty good, mkFit might be giving lower efficiency for pT > 10 GeV.

osschar:
@makortel  your call, I can make my PR right away ... but it probably makes more sense for you to cherry-pick the commit and do the full validation.

makortel:
@osschar Thanks. I'll then cherry-pick it here and provide the full validation.

IHateLinus:
Matti,
Can you please provide the precise definitions of numerator and denominator for these as now that we have two types of seeds its not clear to me if this is for 3+4 hitters or just 3 etc..
Thanks, Avi

makortel:
Both have 3-hit seeds (made from 4-hit pixel tracks by dropping the outermost hit).

IHateLinus:
Thank you!

This makes the pt>10GeV note interesting. I am wondering what this may be telling us.

makortel:
I cherry-picked the commit of @osschar. Here are the standalone benchmarks
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_2/
(IIUC should have no impact)

Here is MTV for 10k ttbar+PU50 events comparing HLT iter0 from CMSSW with HLT iter0 with mkFit
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv/
(with both offline- and online-style MTV cuts). I can make a summary of the comparison for Friday.

slava77:
> This makes the pt>10GeV note interesting. I am wondering what this may be telling us.

this actually looks quite depressing.
Single percent efficiency drop is usually a big deal. Here we have 20% by 100GeV.

IHateLinus:
right. The interesting question is why.
Given the size of the effect, one hopes it would not be hard to pin down.

cerati:
maybe it is not the cause, but are we using the duplicate seed removal? we should disable it for triplets.

makortel:
> maybe it is not the cause, but are we using the duplicate seed removal? we should disable it for triplets.

The duplicate seed cleaning is enabled. I can test again disabling it.

makortel:
> > maybe it is not the cause, but are we using the duplicate seed removal? we should disable it for triplets.
> 
> The duplicate seed cleaning is enabled. I can test again disabling it.

I updated the plots in https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv/ and the seed cleaning seems to have no effect on efficiency (black=disabled, red=enabled)
image
It does have small effect on duplicate rate
image

mmasciov:
About the hit selection windows for pixel layers:
- didn't touch barrel:
https://github.com/trackreco/mkFit/blob/devel/Geoms/CMS/make_trk_info.C#L169
- did slightly modify endcaps:
https://github.com/trackreco/mkFit/blame/devel/Geoms/CMS/make_trk_info.C#L291
from li.set_selection_limits(0.01, 0.05, 0.8, 1.6) to li.set_selection_limits(0.01, 0.015, 0.8, 1.6),
where "set_selection_limits" is defined at https://github.com/trackreco/mkFit/blob/devel/TrackerInfo.cc#L17,
q being "z" for barrel and "r" for endcaps.

That is, all was done for pixel layers was to slightly reduce the phi selection window for the pixel endcaps.

Naively, I am not sure whether this could be the reason for the inefficiency at large pT.

mmasciov:
The cuts were based on efficiency for quadruplet seeds.
For positive endcap, our layers are called 18,19,20:
L.18: https://mmasciov.web.cern.ch/mmasciov/n2seeds_l18/ (empty - probably didn't have any non-seed hits here?)
L.19: https://mmasciov.web.cern.ch/mmasciov/n2seeds_l19/
L.20: https://mmasciov.web.cern.ch/mmasciov/n2seeds_l20/
At that time, 0.015 in phi seemed a good choice.

slava77:
@makortel 
do you have MTV plots for single-muon events?
If we see the efficiency drop there, the debugging of the issue will be much easier.

makortel:
> do you have MTV plots for single-muon events?

No, but I can produce them.

In the meanwhile, here are plots with pT>10 GeV cut on the TrackingParticles (in efficiency plots)
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_ptcut10/
The loss appears to be largest in the very forward, |eta| > 2 (blue=CMSSW, red=mkFit without seed cleaning)
image

makortel:
@slava77 
> > do you have MTV plots for single-muon events?
> 
> No, but I can produce them.

Did you mean 10-muon?

IHateLinus:
Yes. we fondly call them single muon events…

slava77:
it looks like we have more strip hits but less strip layers

https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_ptcut10/plots_hlt_hltIter0PFlowCtfWithMaterial/hitsLayers.pdf
Screen Shot 2019-10-14 at 11 43 15 AM

I'm trying to think of an explanation to this. Are the hits and layers counted in the same way?

cerati:
I guess this is because CMSSW uses matched hits, while we don't

osschar:
Can we plot mkFit layer number distributions for various eta regions (or eta bins, if we have enough tracks)?

makortel:
> Can we plot mkFit layer number distributions for various eta regions (or eta bins, if we have enough tracks)?

If someone tells me how to do that.

osschar:
Hmmh, I guess it would be easier for me to run it myself then :)

Is it ok if I take this branch and then knock off the last hit of the quadruplet seed for one of our new standalone samples?

osschar:
I mean ... I don't know how to do it out of the box ,,, but I'll compile our code with ROOT on and home-cook a TH2 during quality validation.

Maybe Kevin knows of a better way :)

makortel:
> Is it ok if I take this branch and then knock off the last hit of the quadruplet seed for one of our new standalone samples?

I'd expect that to be a good-enough approximation (for this particular issue) of what happens at the HLT.

osschar:
I made TH2s of hit layer vs.eta for 4-hit seeds and 3-hit seeds (pT > 10) and subtracted them:
http://xrd-cache-1.t2.ucsd.edu/matevz/tmp/n_hit_delta.png
Apparently something goes wrong on the initial endcap layers ... I guess I'll need to get personal with some of the affected tracks :)

makortel:
Here are the 10mu plots (with the normal MTV efficiency cuts)
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_10mu/
They're probably not super-useful as the upper pT limit for the muons in our sample is 10 GeV. Anyway, the efficiency starts to decrease already around 6 GeV
image
Efficiency vs. eta has a large (almost 10 % relative) drop around eta 0 in addition to a large drop beyond |eta|>2.4 (efficiency vs pT and others have a cut |eta| < 2.4)
image

kmcdermo:
> I mean ... I don't know how to do it out of the box ,,, but I'll compile our code with ROOT on and home-cook a TH2 during quality validation.
> 
> Maybe Kevin knows of a better way :)

Sorry for the delay in response. As it has already been figured out, I figured I would still share how to do this with the validation trees for posterity. This assumes that the ./val_scripts/validation-cmssw-benchmarks.sh has been run, and the CE SIMVAL(SEED) root file is open in ROOT.

If we just want to plot the number of found mkFit layers vs eta, for every seed, regardless of being matched, one could do the following:
- frtree->Draw("nLayers_build:eta_seed","","colz")

To draw matched, unmatched, etc., add the following selection:
- "mcmask_build==1" for reco tracks matched to sim tracks
- "mcmask_build==0" for reco tracks completely unmatched to sim tracks
- "mcmask_build==2" for reco tracks matched to sim tracks, but the sim track is actually unfindable

It might be worth setting the canvas to Logz(), and also setting the bins to something like:
- ">>hist(30,-3,3,30,0,30)"

osschar:
I spent basically whole day setting up debug printouts and staring at them for 2 tracks, now they are both very special to me :)

The verdict: we pick up crap on the additional pixel layers and with N__max_cands = 6 the good tracks get dropped pretty early. It's the scenario I've been worried about for a while (and asked a couple of times to investigate it): a wrong hit is picked up and errors are blown up so chi2 of those tracks becomes low and so the good tracks get squeezed out.

Anyway, for both of these guys there were two remedies:
1. Increase N_max_cands to 12 (in Geoms/CMS-2017.cc); or
2. Tighten up pixel layer min/max q search windows (this goes in Geoms/CMS-2017.acc). I just put in significantly tighter values (min 0.1, max 0.8, it was (0.8/1.6)) and it worked for both of these tracks, with N_max_cands = 6.

Clearly we need to fine tune this further :)

@makortel I'll make a PR with this changes to your branch here. Please do feel free to modify the values with more informed decision process ;)

osschar:
Oh, apparently I'm allowed to push to your branch ... so ... here you go :)

slava77:
> The verdict: we pick up crap on the additional pixel layers and with N__max_cands = 6 the good tracks get dropped pretty early. It's the scenario I've been worried about for a while (and asked a couple of times to investigate it): a wrong hit is picked up and errors are blown up so chi2 of those tracks becomes low and so the good tracks get squeezed out.

I'm not sure how this aligns with the efficiency drop in the muon gun events as well. There should be no "stray" hits there.

makortel:
Thanks @osschar, I'll run the HLT jobs again (for both ttbar+PU and 10mu)

osschar:
@makortel Can you please clarify how you obtain and process the triplet seeds?
- What are the possible seed layers and possible locations of "holes"?
- If you also do some hit removal ... do you refit the seeds after the hit removal and before passing them to mkFit?

makortel:
>  Can you please clarify how you obtain and process the triplet seeds?
> 
> * What are the possible seed layers and possible locations of "holes"?

In this specific case the layers are always the three innermost one, see
https://github.com/cms-sw/cmssw/blob/1b741f686382ea5f3792ce50e8afa8b27df97e42/RecoTracker/TkSeedGenerator/plugins/SeedGeneratorFromProtoTracksEDProducer.cc#L147-L148

(in general, e.g. in HLT iter2,  the triplets contain all combinations of the 4 layers)

> * If you also do some hit removal ... do you refit the seeds after the hit removal and before passing them to mkFit?

Contrary to what I said last Friday, the `SeedFromConsecutiveHitsCreator` (used by `SeedGeneratorFromProtoTracksEDProducer` when `useProtoTrackKinematics` is `false`) does a fit on the hits (in the same way the 4-hit seeds are fitted in offline initialStep).

https://github.com/cms-sw/cmssw/blob/master/RecoTracker/TkSeedGenerator/plugins/SeedFromConsecutiveHitsCreator.cc

makortel:
Here are MTV plots (with default efficiency cuts) showing the effect of the last commit https://github.com/trackreco/mkFit/pull/242/commits/b027cfe3bb54f8274c1986438656948ae58fd97e
* ttbar+PU50 https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_v2/
* 10mu https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_10mu_v2/

In short, the effect in efficiency is small, and in the wrong direction (plots for ttbar+PU; blue=CMSSW, red=`b027cfe^`, black=https://github.com/trackreco/mkFit/pull/242/commits/b027cfe3bb54f8274c1986438656948ae58fd97e)
image
image
Fake rate increases
image
With https://github.com/trackreco/mkFit/pull/242/commits/b027cfe3bb54f8274c1986438656948ae58fd97e mkFit collects significantly less hits from FPix
image

osschar:
Rats ... I guess I have to find new track friends now :)

osschar:
Just for completeness, can you please undo this change and try increasing max_cand and max_holes like this:
```
matevz@phi3 store-hits-per-seed> git diff Geoms/CMS-2017.cc
diff --git a/Geoms/CMS-2017.cc b/Geoms/CMS-2017.cc
index c3af8a0..d3c53f6 100644
--- a/Geoms/CMS-2017.cc
+++ b/Geoms/CMS-2017.cc
@@ -17,8 +17,8 @@ namespace

     Config::useCMSGeom       = true;
     Config::nlayers_per_seed = 4;
-    Config::maxCandsPerSeed  = 6;  // GC said 3 is enough ???
-    Config::maxHolesPerCand  = 3; // = 12;
+    Config::maxCandsPerSeed  = 12;  // GC said 3 is enough ???
+    Config::maxHolesPerCand  = 4; // = 12;
     Config::chi2Cut          = 15.0; // = 30.0;

     Config::finding_requires_propagation_to_hit_pos = true;
```

makortel:
> Just for completeness, can you please undo this change and try increasing max_cand and max_holes like this:
> 
> ```
> matevz@phi3 store-hits-per-seed> git diff Geoms/CMS-2017.cc
> diff --git a/Geoms/CMS-2017.cc b/Geoms/CMS-2017.cc
> index c3af8a0..d3c53f6 100644
> --- a/Geoms/CMS-2017.cc
> +++ b/Geoms/CMS-2017.cc
> @@ -17,8 +17,8 @@ namespace
> 
>      Config::useCMSGeom       = true;
>      Config::nlayers_per_seed = 4;
> -    Config::maxCandsPerSeed  = 6;  // GC said 3 is enough ???
> -    Config::maxHolesPerCand  = 3; // = 12;
> +    Config::maxCandsPerSeed  = 12;  // GC said 3 is enough ???
> +    Config::maxHolesPerCand  = 4; // = 12;
>      Config::chi2Cut          = 15.0; // = 30.0;
> 
>      Config::finding_requires_propagation_to_hit_pos = true;
> ```

Tried that, results with ttbar+PU50 here
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_v2/
(blue=CMSSW, red=mkFit before, black=mkFit with that patch)
The patch has a rather small effect
image
image
image

osschar:
Thank you! I can assure you it had a substantial effect on running time ;)

mkFit/MkBuilder.cc
```diff
@@ -1532,7 +1534,12 @@ void MkBuilder::PrepareSeeds()
   //0 = not set; 1 = high pT central seeds; 2 = low pT endcap seeds; 3 = all other seeds
   assign_seedtype_forranking();
 
-  fit_seeds();
+  // Do not refit cmssw seeds (this if was nested in fit_one_seed_set() until now).
+  // Eventually we can add force-refit option.
+  if (Config::seedInput != cmsswSeeds)
+  {
+    fit_seeds();
+  }
```
slava77, line 1555: I'm curious if reverting this will fix the efficiency? the old code was pushing things in and out, unclear if there was something useful there that's not done already in (what remains now) just the import_seeds

osschar:
If you revert this, you are back to the crash with wrong number of hits in the seed :)

I'd have to fix the fitting code to allow fitting of cmssw triplet seeds. We actually never fitted cmssw hits ourselves, just the ifs were in the wrong places so everything got prepared for fitting (loading of seed hits) but then just the fit wasn't done. The crash with triplets was because the fitting code was expecting (and is still expecting) 4 hits.

makortel:
@osschar Please find 1k-event 10mu HLT bin file from
`/data2/mkortela/hlt_bin/10_4_0_patch1_10mu_hlt_ccc_all_rec.bin`

makortel:
I should note that in order to process the 10mu sample with the HLT seeds, I had to add protection
```patch
 void MkBuilder::find_duplicates(TrackVec& tracks)
 {
   const auto ntracks = tracks.size();
+  if (ntracks == 0) {
+    return;
+  }
   for (auto itrack = 0U; itrack < ntracks-1; itrack++)
   {
```
to
https://github.com/trackreco/mkFit/blob/c41a0f23814dc7f3d72af1f9d40d8eb795e2bb9d/mkFit/MkBuilder.cc#L1279-L1283
Is that the way to go or is there a better place to protect `find_duplicates` for 0-track events?

areinsvo:
That's probably the best place for it. Otherwise you would have to copy-paste the protection everywhere find_duplicates is called.

makortel:
Thanks @areinsvo. I included the change as a commit here (even if technically it is unrelated and could go alone; if anyone thinks it should go in separately please scream and I'll do that) and removed the commit from @osschar that turned out to not help (I have a local copy of in case still needed).

osschar:
OK, thanks, I'm trying to find myself some new friends :)

osschar:
Oh, you didn't put CMSSW rec-tracks in. Will try by comparing 3 and 4-hit seeds ... but it might be more informative to compare against cms tracks.

makortel:
> Oh, you didn't put CMSSW rec-tracks in. Will try by comparing 3 and 4-hit seeds ... but it might be more informative to compare against cms tracks.

Hmh. I did include `--write-rec-tracks` for `writeMemoryFile`.

osschar:
Yes, I see the flag in the file header ... but the number of rec tracks is always zero. Hmmh, maybe I messed up Event loading while talking to my friends ... but I don't think I touched rec-tracks.

makortel:
Seems that `writeMemoryFile` takes only `initialStep` seeds
https://github.com/trackreco/mkFit/blob/c41a0f23814dc7f3d72af1f9d40d8eb795e2bb9d/tkNtuple/WriteMemoryFile.cc#L747
which of course does not work with the HLT tracks. I'll make a new bin file after an hour or so.

osschar:
Right, there are no seeds, either :)

makortel:
@osschar Could you try again now (same file)?

osschar:
rec-tracks are in ... but still no seeds.

makortel:
I missed one initialStep-requirement. Could you try again (same file)?

osschar:
Yup, now I have everything, thanks! :)

osschar:
I went over comparing seed/sim/cmssw/mkfit tracks of first 20 events (about 100 seeds). I found 3 tracks for which mkfit only finds about 50% of hits (cms finds about 75%) and will look into those in detail. Then there are a couple more that are not so dramatic and I might look into later. All in all, this is about 10 tracks out of 100, so a 10% efficiency drop which seems consistent within +-10% error bar here :)

Another thing I noticed is that about 1% of seeds get rejected in nan_and_silly check because they have negative values on element (5,5) of covariance (theta), e.g.:
```
Post-cleaning seed silly value check and fix (label=1): (5,5)=-7.445001e-07
Post-cleaning seed silly value check and fix (label=9): (5,5)=-5.498798e-08
Post-cleaning seed silly value check and fix (label=5): (5,5)=-2.132469e-08
Post-cleaning seed silly value check and fix (label=5): (5,5)=-9.654286e-10
Post-cleaning seed silly value check and fix (label=5): (5,5)=-5.437179e-06
```
There are also two cases with similar values on (4,4) (phi). pT is also rather poor but the error is not negative.

Right now our settings are to discard such seeds (`Config::nan_n_silly_remove_bad_seeds = true`). We have an option implemented to fix negative values on the diagonal (`Config::nan_n_silly_fixup_bad_seeds`) but it is off right now. Also, it sets the value to 1 ... there might be a better choice for this, probably depending on which parameter is affected.

Do we want to go into this business tight now? It's a 1% effect at most.

osschar:
I went through one track, pT=8.49, eta=-2.39

To get mkfit to find a track of about the same quality as cmssw (10 found hits, pT 8.25, ch2 ~ 60) I had to:
- increase min_dphi search window x3 for the endcap layers
- increase max hit chi2 from 15 to 20

And still, a short track with 3 seed hits and a wrong hit picked on the 4th layer, followed by three missed hits wins at the end, mostly because of the chi2 penalty to candidate score (the "good" track has 6 more hits but chi2 of 60 makes it worse). Clearly, our scoring isn't optimal for noisy 3-hit seeds :)

It also seems our detection of hit/miss of a layer is not fuzzy enough ... we might have to consider extending the might-have-missed-the-layer region or increasing N_max_holes.

I will look at two more tracks tomorrow. Then I might have a recommendation for a couple of test runs to see how this shows on a larger statistics.

slava77:
> And still, a short track with 3 seed hits and a wrong hit picked on the 4th layer

was this still from the muon gun sample?
I'm curious where is the wrong hit from. Does it have an MC info?

osschar:
Yes, muons. i was able to trace it to another track. You can get full event dump by uncommenting all the DUMP_XXX defines in Event.cc:456 and run
```
mkFit/mkFit --cmssw-n2seeds --input-file /data2/mkortela/hlt_bin/10_4_0_patch1_10mu_hlt_ccc_all_rec.bin --read-cmssw-tracks --build-std --quality-val --num-events 1 --start-event 12
```
and then the hit that gets picked up is the only hit on layer 54, belogning to track label 1. The track that I was looking at had label 6.

osschar:
I looked at three more tracks, here's a short report on what I had to fix to improve mkFit track.
[ Format note: mkFit best track | sim track, CMS track in next line ]

1. 19 7 4/8 chi2=0.284564 pT=10.207181 eta=2.411617  |  15/15 chi2=12.000000 pT=8.312447 eta=2.411341
  4 with q=-1 pT=  8.143 eta=  2.411 nHits=12  label=   4
- Increase chi2 per hit cut from 15 -> 20 to pick up the correct hit on on inner layer

2. 3 6 13/17 chi2=12.091969 pT=7.028310 eta=-1.857068  |  23/23 chi2=16.000000 pT=7.007823 eta=-1.858734
  4 with q=+1 pT=  7.011 eta= -1.858 nHits=18  label=   4
- Increase N_max_holes from 3 -> 4, apparently layer r/z limits are to "rough", at least in the endcap.

3. 14 1 5/9 chi2=5.698732 pT=21.901735 eta=-2.483779  |  11/11 chi2=7.000000 pT=9.415405 eta=-2.481886
  4 with q=-1 pT=  9.475 eta= -2.483 nHits= 7  label=   4
- chi2 -> 30 (28.7)
- increase N_max_holes
- reduce chi2 penalty or increase hit bonus
- a rather hopeless case, note that cmssw also only found 7 hits out of 11.

So, there isn't a single thing that's wrong but rather our whole scoring and best-short-track handling seems to be optimized for 4-hit seeds with better initial parameters.

We need to figure out how to handle this but it seems we need:
- a good way to automatically tune the track scoring parameters / function,
- have different config parameter sets for different iterations (or they could be set per seed as well, based on N_hits, pT, eta, initial track parameter errors).

slava77:
> 19 7 4/8 chi2=0.284564 pT=10.207181 eta=2.411617 | 15/15 chi2=12.000000 pT=8.312447 eta=2.411341

I'm not sure I understand the numbers before chi2. Please clarify what is printed.
What are the seed track parameters in all these cases?

osschar:
EventNo Label hits_found/total_hits
Seeds:
1.    4 q=-1 pT= 10.979 eta=  2.411 nHits=3 label= 7
2.    4 q=+1 pT=  5.916 eta= -1.858 nHits=3 label= 6
3.    4 q=-1 pT=  6.122 eta= -2.482 nHits=3 label= 1

slava77:
> EventNo Label hits_found/total_hits

is the difference between hits_found and total_hits only in the number of holes between the innermost and the outermost found hit?

makortel:
Here are MTV plots comparing CMSSW with triplet seeds (blue), mkFit with quadruplet seeds (red), and mkFit with triplet seeds (black) (and to remind, in this specific case the triplet seeds contain the first three hits of the quadruplet seeds)
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad/

image

IHateLinus:
So sadly, we seem to suck at every level above 10 GeV…

Even our quads seem worse at high pT than CMSSW w only three pixel hits used in the seed. This means that CMSSW manages to pick the 4th pixel hit and/or extend the track correctly based on the three inner most hits while we don’t.

Moreover, even when we do have them, we still manage to get confused at high(er) pT.

This may again hint at errors shrinking and possible chi2 exploding for high pT tracks?

Does this make sense??

osschar:
> > EventNo Label hits_found/total_hits
> 
> is the difference between hits_found and total_hits only in the number of holes between the innermost and the outermost found hit?

No, it's all the not found hits, including the trailing ones. In this case this just tells you the best selected track was the first short track we found, with 4 not found hits at the end.

cerati:
One issue we can have at high pT is that our track parameters does not allow to change the charge, while CMSSW allows it (Matti just verified it). So if the seed has the wrong charge we are lost. This is something we should fix in the code (remove the charge as a standalone parameter and get it is the sign of the 1/pt parameter, now q/pt).

IHateLinus:
Interesting!!
Can we check for the size of this effect by counting the tracks found in cmssw w opposite sign to seed?

cerati:
we know that on the 10 mu sample (0.5-10 GeV) it ~1% overall but ~8% when looking only at pT>7 GeV and |eta|>2.3. We are now trying to assess the effect on the efficiency, but may need a larger sample to get final conclusions...

cerati:
("we" are FNAL+Slava having a small f2f instead of beer on Friday)

IHateLinus:
Thanks!

makortel:
We had some discussion whether we had seen the high-pT inefficiency before (https://github.com/trackreco/mkFit/pull/242#issuecomment-546470741, red (mkFit with quads). vs blue (CMSSW with triplets)
image
)

I ran both CMSSW and mkFit in the offline initialStep with the full 10kevent ttbar+PU50 sample, MTV plots here
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_offline/
and the efficiency vs. pT below
image

There is some inefficiency in mkFit vs. CMSSW starting around 80 GeV, but the effect is smaller than in the HLT case.

Thinking out loud, one possible cause for the difference is the duplicate cleaning within pixel tracking. I could try what happens in the HLT case for mkFit if I disable that.

makortel:
> Thinking out loud, one possible cause for the difference is the duplicate cleaning within pixel tracking. I could try what happens in the HLT case for mkFit if I disable that.

Plots are here ~~https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad_nopixeltrackclean/~~
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad_nopixeltrackclean3/
* blue: CMSSW with triplet seeds
* red: mkFit with quadruplet seeds
* black: mkFit with quadruplet seeds with CMSSW-pixel-tracking-side duplicate cleaning disabled

Disabling the pixel track cleaning improves the efficiency closer to offline quadruplets, but at the cost of increasing fakes and duplicates. (mkFit's seed cleaning was enabled for both red and black)

cerati:
Hi Matti, I have a hard time reconciling the efficiency plot vs pT with the one vs eta for HLT:
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad_nopixeltrackclean/plots_hltOnline_hltIter0PFlowTrackSelectionHighPurity/effandfakePtEtaPhi.pdf
This has nothing to do with the black dots.
Looking at the one vs eta we have the same or better performance, while vs pT we do significantly worse (not just for pT<0.9 which is not included in the eta plot). I suspect there are different cuts in eta for the plot vs pT (it seems to me there are no blue points beyond |eta|=2.5).

makortel:
> Looking at the one vs eta we have the same or better performance, while vs pT we do significantly worse (not just for pT<0.9 which is not included in the eta plot). I suspect there are different cuts in eta for the plot vs pT (it seems to me there are no blue points beyond |eta|=2.5).

Thanks @cerati for pointing it out, there is clearly something wrong in the blue reference. I re-ran it and made a new set of plots
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad_nopixeltrackclean3/
that look much better.
Legend is the same
* blue: CMSSW with triplet seeds
* red: mkFit with quadruplet seeds
* black: mkFit with quadruplet seeds with CMSSW-pixel-tracking-side duplicate cleaning disabled

cerati:
Great, thanks Matti!
Looking at
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_mtv_quad_nopixeltrackclean3/plots_hltOnline_hltIter0PFlowCtfWithMaterial/effandfakePtEtaPhi.pdf
on average (plot vs phi) red has slightly higher efficiency and slightly lower fake+duplicate rate than blue; the remaining issue is some efficiency loss for pT>=50 GeV.

makortel:
Rebased on top of HEAD of `devel` (merge of #247), and updated README for two new options for the HLT-tracking configuration (when ran in CMSSW). Standalone benchmarks are here
https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR242_3/

PR #250, opened 2020-01-23
Fix negative values in seed covariance matrices

areinsvo:
This PR improves the efficiency of building high pt (> 100 GeV) tracks by manually fixing negative values in the covariance matrix of the seeds. The benchmark plots for the ttbar PU 50 sample can be found here: http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR250. The changes are minimal, since most tracks are < 100 GeV. 

For the high pt 10muon sample with offline seeds (found at /data2/slava77/samples/2018/pass-925bb57/initialStep/default/10muPt0p2to1000HS/memoryFile.fv4.clean.writeAll.CCC1620.recT.191108-c41a0f2.bin), the efficiency jumped from 81.5 to 99.8%, at the expense of an increase in duplicate rate (11.4% before, 18.4% after). See plots [before](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10mu_master/SIMVAL_MTV_SEED/) and after [after](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10mu_fixupBadSeeds/SIMVAL_MTV_SEED/). 

For the high pt 10 muon sample with HLT triplet seeds (found at /data2/slava77/samples/2018/pass-f2bb882/hltIter0/default/triplet/10muPt0p2to1000HS/memoryFile.fv4.clean.writeAll.CCC1620.recT.200115-0a18240.bin), the efficiency jumped from 76.7% [before](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10muHLT_master/SIMVAL_MTV_SEED) to 99.6% [after](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10muHLT_fixupBadSeeds/SIMVAL_MTV_SEED/).

PR #252, opened 2020-02-05
Fix charge flip

areinsvo:
NEEDS DISCUSSION AND REVIEW
This PR fixes the code to consistently handle the case when the track pt goes negative, corresponding to a change in the charge hypothesis for the particle. This has been discussed at two different mkFit meetings: [Dec 13, 2020](https://indico.cern.ch/event/845069/contributions/3548502/attachments/1962289/3261785/ChargeAssignment.pdf) and [Dec 20, 2020 ](https://indico.cern.ch/event/845070/contributions/3548517/attachments/1965389/3267950/ChargeAssignment.pdf). The benchmarks for TTBar PU50 can be seen [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR251/), demonstrating essentially no change . Note that these results are on top of the Config changes introduced in [PR250](https://github.com/trackreco/mkFit/pull/250). 

Someone needs to closely double check my work to make sure I am doing what I intend to and disentangling the sign of ipt from the charge of the particle. There is some evidence that the track quality gets slightly **worse** in the 10muon sample after my changes. See the summary in the [slides posted here](https://indico.cern.ch/event/845077/). 

10 muon results with duplicate removal applied: [Before](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10mu_fixupBadSeeds) and [After](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/PR251_10muOffline)

10 muon results without duplicate removal applied: [Before](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/PR250_10muOffline_noDupRemoval) and [After](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/PR251_10muOffline_noDupRemoval)

### DIFF
--- a/Config.h
+++ b/Config.h
@@ -393,8 +393,8 @@ namespace Config
 
   constexpr bool nan_n_silly_check_seeds      = true;
   constexpr bool nan_n_silly_print_bad_seeds  = false;
-  constexpr bool nan_n_silly_fixup_bad_seeds  = false;
-  constexpr bool nan_n_silly_remove_bad_seeds = true;
+  constexpr bool nan_n_silly_fixup_bad_seeds  = true;
+  constexpr bool nan_n_silly_remove_bad_seeds = false;
 
   constexpr bool nan_n_silly_check_cands_every_layer     = false;
   constexpr bool nan_n_silly_print_bad_cands_every_layer = false;
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -190,7 +190,8 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
 
   const auto assignbins = [&](int itrack, float q, float dq, float phi, float dphi){
 
-    float thisPt    = 1.0f/Par[iI].At(itrack,3,0);
+    //Keep in mind that ipt can be negative, so take abs here
+    float thisPt    = std::fabs( 1.0f/Par[iI].At(itrack,3,0));
     float thisEta   = std::fabs( getEta( Par[iI].At(itrack,5,0) ) );
     //
     float min_dq    = L.min_dq();
--- a/mkFit/PropagationMPlex.cc
+++ b/mkFit/PropagationMPlex.cc
@@ -285,6 +285,8 @@ void helixAtRFromIterativeCCSFullJac(const MPlexLV& inPar, const MPlexQI& inChg,
 	continue;
       }
 
+      //fixme: This function is not currently used, but note that ipt is allowed to take
+      //negative values, which may break the propagation
       const float ipt   = inPar.ConstAt(n, 3, 0);
       const float phiin = inPar.ConstAt(n, 4, 0);
       const float theta = inPar.ConstAt(n, 5, 0);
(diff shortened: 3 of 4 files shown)


slava77:
Is to goal of this PR to fully change the CCS from 1/pt to q/pt ?
Only the parameter vector is modified.
I do not clearly see related changes in the covariance treatment.
Are the changes in the state parameters fully addressing the propagation Jacobian?

areinsvo:
> Is to goal of this PR to fully change the CCS from 1/pt to q/pt ?

No, this PR is a patch that keeps the current solution, where q is a parameter of the track and the propagation functions assume 1/pt is positive. A longer term solution (that would require more work) would be to remove q as a parameter, and instead use q/pt as you suggest.

> I do not clearly see related changes in the covariance treatment.

The issue with the negative values in the covariance matrix was patched in PR250 by simply switching the flags in the Config function to use the "fix" for bad values (setting it to an arbitrary value of 1) rather than removing those seeds. This PR includes those changes as well, but my assumption is that PR250 will be merged first before this one goes in.

>  Are the changes in the state parameters fully addressing the propagation Jacobian?

I believe so, but someone needs to independently review it before we hit the merge button.

slava77:
> No, this PR is a patch that keeps the current solution, where q is a parameter of the track and the propagation functions assume 1/pt is positive. A longer term solution (that would require more work) would be to remove q as a parameter, and instead use q/pt as you suggest.

Does this really work properly?
The Kalman fit supposedly should use a consistent definition of parameters and their covariance.
Do you have some (symbolic) math to show that the current solution is OK?

> The issue with the negative values in the covariance matrix was patched in PR250 by simply switching the flags in the Config function to use the "fix" for bad values

that's not related to my question.

cerati:
Hi Slava,

the Kalman update already allows ipt to become negative, there is no constraint there. The problem we are addressing here is that the propagation instead assumes that ipt positive. The track charge parameter itself is never changed in the code.

So what Allie does is to make sure that when indeed ipt is negative the propagation is done assuming the opposite charge.

makortel:
@cerati What happens if the ipt becomes negative after some layer and stays such? Is the track charge flipped in the end, or does it stay the same as of the seed?

cerati:
the charge parameter is not changed at the moment (we should double check the conversion to CMSSW track is consistent)

areinsvo:
The conversion to CMSSW track takes the charge from the track candidate state, so we need to make sure the final charge flips if necessary: https://github.com/makortel/cmssw/blob/mkfit_1040p1_dev/RecoTracker/MkFit/plugins/MkFitOutputConverter.cc#L410

A check could be added to the cmssw_export function of MkBuilder to update the charge if ipt is negative. Or we could update both as soon as ipt becomes negative, but I'd have to think carefully to make sure that it is done in the right place.

areinsvo:
As mentioned last week and in the slides from December, this PR addressed three places where negative ipt could have unintended consequences: 1) in the propagation 2) when we determine the window sizes for selecting hits 3) in the application of material effects. By removing the fixes to items 2 and 3, I confirmed that basically all of the track quality changes can be traced to the changes for item 1.

I tested this using the high pt 10 mu sample:
If I only fix item 1, and leave item 2 and 3 alone (with duplicates removed) the plots are [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/noFixSelectHitIndices_noFixMaterialEffects/). 
I also checked plots **without duplicate removal applied** if I only fix items 1 and 2 [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/noFixMaterialEffects_noDR/) or if I only fix items 1 and 3 [here](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/noFixToSelectHitIndices_noDR/).

areinsvo:
Closing this pull request, because I implemented a better fix and made a [pull request](https://github.com/trackreco/mkFit/pull/272) to the "Pandora's Box" branch, where we will do a retuning after including several changes such as the charge flip, parametric B field, material effects, and the pickup of overlapping hits.

Issue #261, opened 2020-02-20
update Makefile.config and init-env.sh for gcc8


srlantz:
Dan determined that the following steps are appropriate to run, in order to build mkFit with the gcc8 that is currently defined in init-env.sh:

source /cvmfs/cms.cern.ch/slc7_amd64_gcc820/external/tbb/2019_U9/etc/profile.d/init.sh
export TBB_PREFIX=$TBB_ROOT
make -j24 CXX=g++

The above recipe is needed to supersede the hack location for TBB that Makefile.config sets for any non-icc compiler. (1) The Makefile.config hack is undesirable and should be removed. (2) The correct TBB location should be set in the user environment; thus, the above recipe should be inserted into init-env.sh.

In addition, AVX2:=1 in Makefile.config doesn't work for gcc unless -march=native is also specified; otherwise the appropriate immintrin.h isn't found (see issue #240). This should also be fixed.

dan131riley:
TBB_PREFIX currently overrides the compiler default, so if we set it in init-env.sh we will get TBB from cvmfs even when we're building with icc, instead of the version Intel ships with icc.  Is that what we want?

srlantz:
I see your point, I guess the logic of Makefile.config will also need to be changed so that different things happen depending on icc vs. gcc and whether CMSSW_BASE is defined.

makortel:
`CMSSW_BASE` is not set when CMS builds the externals.

srlantz:
This is fixed in #262, so when that PR is merged, this issue can be closed.

Note, the existing logic in Makefie.config is left intact in PR #262; if the compiler is g++, make will check the environment for TBB_GCC, and if it exists, it will use that as the TBB_PREFIX. Therefore, TBB_GCC should not be set for clang, etc.

The resolution for the AVX2 problem turned out not to depend on the -march=native flag. The reason that flag seemed to be required for AVX2 was due to a flaw in MatriplexCommon.h, where it was actually borrowing an intrinsic from AVX512F. In PR #262, an equivalent AVX2 intrinsic was substituted. The only other change needed was to add -mfma to the compiler options, to comply with gcc's strict association between supplied x86 options and the intrinsics that are included in immintrin.h.

Issue #171, opened 2018-10-11
Makefile.config inconsistencies between icc and gcc


srlantz:
When AVX_512 is defined in the make command line, the compiler options for gcc and icc become, respectively (from Makefile.config):
```
VEC_GCC  := -mavx512f -mavx512cd
VEC_ICC := -xHost -qopt-zmm-usage=high
```
However, these sets of options are not doing the same thing: `-mavx512f -mavx512cd` (gcc) translates to `-xCOMMON-AVX512` (icc), whereas `-xHost` (icc) translates to `-march=native` (gcc). The latter means to use instructions appropriate to the processor that is running the compilation. This actually brings in quite a few more AVX-512 instructions on SKL-SP vs. KNL (i.e., beyond the subset common to both). It's also troubling that on SNB, specifying `make AVX-512:=1` produces code that has no AVX-512 instructions at all.

Relevant comments already appear in PR #165, especially:

- [@dan131riley's comment](https://github.com/cerati/mictest/pull/165#issuecomment-428706470) that `-xCOMMON-AVX512` was intentionally replaced with `-xHost` in PR #140 due to an apparent icc compiler bug (reference to issue #139),  and
- [@kmcdermo's comment](https://github.com/cerati/mictest/pull/165#issuecomment-428669707) that in the benchmarking scripts we don't cross-compile anymore, we scp the source directly to each test machine and compile locally.

Whatever the history, I think we want to keep these options consistent between icc and gcc. There are a couple of ways to do this.

1. The icc compiler bug may be fixed by now. If so, we specify `-xCOMMON-AVX512` again for icc when `AVX_512:=1` and include a comment about which versions of icc have trouble with it.
2. We make a new distinction in Makefile.config, so that icc uses `-xCORE-AVX512` for Skylake-SP and `-xMIC-AVX512` for KNL. Similarly, gcc would use `-march=skylake-avx512` and `-march=knl` in those cases (these options are valid in recent gcc versions). The icc options are equivalent to `-xHost` on each AVX-512 platform, so they should not trigger the compiler bug in icc versions that are affected by it. Accordingly, we would eliminate `AVX_512` from make and split it into (say) `CORE-AVX512` and `MIC-AVX512`.

In addition, we could consider changing the default vectorization options to be `-xHost` (icc) and `-march=native` (gcc). This may be more sensible than what we currently do, which is to specify the lowest common denominator of all the available processors.

srlantz:
[...]

PR #262, opened 2020-02-27
fix options and intrinsics for gcc8 and avx2

srlantz:
This PR addresses issues #261 and #171. The changes are pretty minor and mostly affect gcc compilation. A single AVX2 intrinsic function was changed in MatriplexCommon.h, as the one previously used was actually borrowed from AVX-512, and gcc is extremely strict about which immintrin.h is included for a given set of x86 options. Also the -mfma flag was added for AVX2, as it turns out that the FMA instructions are not automatically part of the AVX2 ISA extension.

I didn't run the usual benchmarks because none of these changes touches our usual benchmark build, which relies on icc and AVX-512.

However, in the course of testing, I noticed that the number of builtcands would vary slightly depending on compiler and AVX level. I'll open a separate issue about that, because the issue appears to predate the changes in this PR (it was happening as far back as PR #229).

### DIFF
--- a/Makefile.config
+++ b/Makefile.config
@@ -72,11 +72,11 @@ OPT := -g -O3
 
 # 4. Vectorization settings
 ifdef AVX_512
-VEC_GCC  := -mavx512f -mavx512cd # -march=native -fopt-info-vec -mavx
-VEC_ICC  := -xHost -qopt-zmm-usage=high #-march=native -mtune=native #-xcore-avx512
+VEC_GCC  := -march=native # -fopt-info-vec -mavx512f -mavx512cd
+VEC_ICC  := -xHost -qopt-zmm-usage=high # -xcore-avx512
 else ifdef AVX2
-VEC_GCC  := -mavx2
-VEC_ICC  := -mavx2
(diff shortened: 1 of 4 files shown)


Makefile.config
```diff
@@ -201,17 +201,16 @@ ifneq ($(CXX),icc)
...
+      TBB_PREFIX = $(TBB_GCC)
     endif
   endif
 endif
 
 ifdef WITH_TBB
   # icc finds tbb in its own installation, but allow overriding in case it doesn't
   ifdef TBB_PREFIX
-    CPPFLAGS += -I${TBB_PREFIX}/include
+    CPPFLAGS += -I${TBB_PREFIX}/include/tbb
```
makortel: Why is `/tbb` needed here? Our `#include`s are already like `#include <tbb/tbb.h>`, no?

srlantz: Doesn't seem to make a difference either way, but you're right, I'll fix this

srlantz:
I'll run the benchmarks for this PR if I get even one request to do so (I realize it's the protocol)

PR #265, opened 2020-03-02
Fix backward fit for tracks without a hit on innermost pixel layer

makortel:
Otherwise triplet seeds without hits on BPix1 will lead to segfaults.

The fix itself came from @osschar.


makortel:
Benchmarks can be found from https://mkortela.web.cern.ch/mkortela/tracking/mkfit/PR/PR265/

PR #268, opened 2020-03-12
remove usolids

davidlange6:
usolids is an obsolete library, and seems not used in mkfit.


Makefile.config
```diff
@@ -31,7 +31,7 @@
 
 # 1. Use ROOT or not (never used on MIC)
 # Comment out to disable root ("1" is not relevant)
-#WITH_ROOT := 1
+WITH_ROOT := 1
```
dan131riley: @davidlange6 please revert this, ROOT defaults shouldn't be changed

davidlange6: whoops - my mistake

PR #269, opened 2020-03-12
fix scale param to mask_i32gather to be a compile time constant

dan131riley:
The "scale" parameter to the mask_i32gather intrinsics is supposed to be a compile time constant.  icc with optimization lets this get by, but with optimization off it doesn't, and clang catches the problem regardless of optimization level.  This change templates the intrinsics versions of SlurpIn so that we can use type deduction to get a constexpr scale.  This change should have *no* physics or timing consequences, it only serves to make compilation possible with picky compilers, so I'm not planning on running validation.

### DIFF
--- a/Matriplex/Matriplex.h
+++ b/Matriplex/Matriplex.h
@@ -100,7 +100,8 @@ public:
 
 #if defined(MIC_INTRINSICS)
 
-   void SlurpIn(const T *arr, __m512i& vi, int scale, const int N_proc = N)
+   template<typename U>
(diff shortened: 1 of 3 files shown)


srlantz:
These changes just move a few sizeof() operations from the caller to the callee. Given that the calling routines in MatriplexPackers.h are already templated with data type D (=float, pretty much), it makes a lot of sense to template the intrinsics-based variants of SlurpIn in the same way. I guess C++ compilers will be smart enough not to construct the nameless D() argument and put it on the stack for SlurpIn, when all the SlurpIn template function needs to know is sizeof(D). Merging.

Issue #251, opened 2020-01-23
Duplicate rate increase at high pt


areinsvo:
This GH issue is a followup issue to [issue #248](https://github.com/trackreco/mkFit/issues/248). When studying the efficiency loss at high pt, it was discovered that fixing the negative entries in the covariance matrix led to an increase in the duplicate rate. More generally, using the high pt 10μ sample indicated that the duplicate rate grows dramatically [with pt](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10mu_fixupBadSeeds/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_dr_pt_build_pt0p0_SIMVALSEED.png) and at [high eta](http://areinsvo.web.cern.ch/areinsvo/MkFit/EffHighPt_Jan2020/testHighPt10mu_fixupBadSeeds/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_dr_eta_build_pt0p0_SIMVALSEED.png). For pt < 100 GeV the duplicate rate is ~2% and for pt > 500 the duplicate rate is 20%. We might need a high-pt retuning of the duplicate removal parameters.

Musings from Matevz regarding the increase after the covariance matrix fix [PR250](https://github.com/trackreco/mkFit/pull/250): There is a possibility that we have a fit that is poorly constrained and so the result is somewhat "random". If this is indeed the case, seeds from the set of hits belonging to the same mc-track would lead to drastically different parameters and there's no way to clean those out on track parameter consideration.

PR #270, opened 2020-03-27
Duplicate Removal

tresreid:
Responding to issue #251  
Changed the pt cut in duplicate removal from |pt2-pt1|/maxpt to |1/pt2 -1/pt1|
changed the maxdPhi cut to 0.5 and the maxdPt cut to 0.5
Benchmarks for ttbar can be found here: http://mireid.web.cern.ch/mireid/mkFit_keep/duplicate_rm/benchmarks_duplicateRemovalPR_ttbar/
Benchmarks for muons can be found here: 
http://mireid.web.cern.ch/mireid/mkFit_keep/duplicate_rm/benchmarks_duplicateRemovalPR_muons/

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -87,8 +87,8 @@ namespace Config
 
   bool  removeDuplicates = false;
   bool  useHitsForDuplicates = true;
-  float maxdPhi = 0.4;
-  float maxdPt  = 0.7;
+  float maxdPhi = 0.5;
+  float maxdPt  = 0.5;
   float maxdEta = 0.05;
   float minFracHitsShared = 0.75;
   float maxdRSquared = 0.000001; //corresponds to maxdR of 0.001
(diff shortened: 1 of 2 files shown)


areinsvo:
Thanks @tresreid. The new plots look really good, especially the duplicate rate with respect to pt for the 10muon sample. There are no noticeable changes to timing. If there are no objections, I plan to hit the merge button for this PR tomorrow.

makortel:
Is the origin of the duplicates in the 10mu sample understood? (i.e. from additional seeds per particle because of module overlaps, or is there something fancier going on)

tresreid:
I've tried looking at how some of the different hits affect the duplicate rates. Some examples can be found in the backup slides here: https://indico.cern.ch/event/895996/contributions/3777784/attachments/2010816/3359978/Duplicate_removal_v2.pdf
I think for the most part different hits in the seeds or extra hits at the end can cause duplicates (or rather a large change in the pt (or 1/pt)).

PR #275, opened 2020-06-11
Overlaps and fix for low-pt FP exceptions

osschar:
* Pickup of overlap hits during outward search

  - Overlaps are stored as regular hits.
  - Config variables for chi2 cuts and min pT for adding an overlap:
    chi2CutOverlap pTCutOverlap. Defailts are 3.5 and 1.0.
  - Config variable for overlapHitBonus -- set to 0 so overlap
    hit is scored as a normal hit.

* A major source of NANs was low-pT tracks that could not reach a barrel layer
  during PropagateToR.

  The fix:

    - Detect in low-level propagate-to-r if dot product between momentum and
      radial direction is below 0.2 (~80 degrees) and set the propagation failed
      flag.
    - The wrapper propagate-to-r checks for this and copies input parameters
      over the output.
    - Later, in MkBuilder::handle_missed_layers(), the radius is checked agains
      the min_r of the layer and the track is stopped (for barrel) or handled as
      if it has missed the layer (for transition region where furhgter endcap
      layers could still be reached).

  Open issues:

  - This could be implemented otherwise, either by checking radius in
    SelectHitIndices or through passing out failed propagation flag out of the
    PropagateXyz() functions.

  - BestHit does not stop the tracks or even check if the layer has been
    missed. If the check goes into SelectHitIndices it could.

* Try to trace down other sources of FP exceptions. Found some, couldn't fix
  all.

  - Track::swimPhiToR() asin(x) where x > 1 or x < -1.
    // XXX-ASSUMPTION-ERROR can not always reach R, should see what callers expect.
    // For now return PI to signal apex on the ohter side of the helix.

  - MkFinder::SelectHitIndices() sqrt of err(2,2) which turns negative.
    Left an abs there and comment:
    // XXX-NUM-ERR above, Err(2,2) gets negative!

  - MkFinder::FindCandidatesCloneEngine() negative chi2
    Left an abs there and comment:
    // XXX-NUM-ERR assert(chi2 >= 0);


osschar:
Validation plots:
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/275-Overlaps-lowPtExc/

osschar:
And validation after BestHit track stopping change:
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/668-besthit-stopping-after/

areinsvo:
@osschar after updating PandorasBox, there are now merge conflicts with this PR. Can you take a look?

osschar:
I don't know if you got the notification ... I resolved the conflicts through github interface, amazing stuff :)

areinsvo:
Thanks @osschar!

PR #276, opened 2020-07-15
support for iterations: seeds

slava77:
- TrackAlgorithm scoped enum is added (in Config.h) 
- TrackBase::Status is extended with 6 bits to save the algorithm value
- WriteMemoryFile is modified to
    - be able to write all available seeds
    - to save the algorithm flag for the seeds and tracks
    - to save seed hits including the strip detector: if "glued" hits are used in seeding but the config (default) is not asking to use matched hits, these hits are split into their mono and stereo components

bin files with all seeds saved are available in phi2:/data2/slava77/analysis/CMSSW_10_4_0_patch1_mkFit/pass-2eaa1f7/initialStep/default/
for muon guns (dirs 10muPt0p2to10HS and 10muPt0p2to1000HS) and for ttbar (dir 11024.0_TTbar_13; for PU 0 and PU50 )
- look for trackingNtuple.root files for the original ntuples
- look for memoryFile.fv4.clean.writeAll.CCC1620.recT.allSeeds.200714-02cb125.bin for the binary files
- NB: PU50 is still being made as of the original posting

The "allSeeds" binary files are at this point only useful for studies of the seed properties.

### DIFF
--- a/Track.h
+++ b/Track.h
@@ -231,8 +231,11 @@ public:
 	// Whether or not the track matched to another track and had the lower cand score
 	bool duplicate : 1;
 
+        // Tracking iteration/algorithm
+        unsigned int algorithm : 6;
+
         // The remaining bits.
-        unsigned int _free_bits_ : 25;
+        unsigned int _free_bits_ : 19;
 
       };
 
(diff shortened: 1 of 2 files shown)


osschar:
Yes, as discussed, the enum class TrackAlgorithm fits better into Track.h (or a new file [Cms]TrackAlgorithm.h). Otherwise it looks good. Do you want to merge as is and move it later or do the placement before?

slava77:
> Do you want to merge as is and move it later or do the placement before?

I'd move first.

slava77:
> Yes, as discussed, the enum class TrackAlgorithm fits better into Track.h (or a new file [Cms]TrackAlgorithm.h). Otherwise it looks good. Do you want to merge as is and move it later or do the placement before?

I moved the enum to Track.h
Now this is a part of the TrackBase class in the same way as in CMSSW, perhaps it's more appropriate than to decouple it standalone.

PR #282, opened 2020-10-23
add hit masks per hit to mark hits available for specific iterations

slava77:
The purpose is to pass the information about hits allowed/blocked to be used in a specific iteration to `mkfit::Event`.

- `layerHitMasks_` of type `vector<vector<uint64_t> >` is added to `mkfit::Event`, it should be aligned with `layerHits_` and provides a mask of bits which show with 0/1 if a hit is allowed/blocked to be used in a given iteration. The bit order is the same as `TrackBase::TrackAlgorithm`
    - file format version is updated to 5; reading of version 4 is supported
    - unless `layerHitMasks_` are read from a file via `Event::read_in`, for other Event initialization via `read_in` and `Simulate` the masks are filled with 0 (all hits unmasked)
- WriteMemoryFile.cc can write the masks to the file if enabled with `--write-hit-iter-masks`

I did not run the validation yet; no changes are expected at this point. Still, the following better be tested first
- [x] reading the old v4 files
    - timing looks the same between baseline as of #279 http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/val/phi3/devel-pr279/benchmarks/MultEvInFlight/SKL-SP_CMSSW_TTbar_PU50_CE_MEIF_time.png and this PR as of 
3c860d5  http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/val/phi3/devel-pr279-iters/3c860d5/benchmarks/MultEvInFlight/SKL-SP_CMSSW_TTbar_PU50_CE_MEIF_time.png
    - the efficiency is identical as expected http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/val/phi3/devel-pr279/benchmarks/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_eff_eta_build_pt0p0_SIMVALSEED.png in baseline vs http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/val/phi3/devel-pr279-iters/3c860d5/benchmarks/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_eff_eta_build_pt0p0_SIMVALSEED.png in this PR as of 3c860d5
- [ ] reading the new files

@mmasciov I hope that the changes here (as of PR open, at least) are small enough to not interfere with your setup for using multiple iterations.

### DIFF
--- a/Event.cc
+++ b/Event.cc
@@ -57,6 +57,7 @@ Event::Event(int evtID) :
 {
   reset_nan_n_silly_counters();
   layerHits_.resize(Config::nTotalLayers);
+  layerHitMasks_.resize(Config::nTotalLayers);
 }
 
 Event::Event(const Geometry& g, Validation& v, int evtID, int threads) :
@@ -65,6 +66,7 @@ Event::Event(const Geometry& g, Validation& v, int evtID, int threads) :
 {
   reset_nan_n_silly_counters();
   layerHits_.resize(Config::nTotalLayers);
+  layerHitMasks_.resize(Config::nTotalLayers);
 
   validation_.resetValidationMaps(); // need to reset maps for every event.
 }
@@ -497,6 +499,7 @@ void Event::read_in(DataFile &data_file, FILE *in_fp)
   int nl;
   fread(&nl, sizeof(int), 1, fp);
   layerHits_.resize(nl);
+  layerHitMasks_.resize(nl);
   for (int il = 0; il<nl; ++il) {
     int nh;
     fread(&nh, sizeof(int), 1, fp);
--- a/Event.cc
+++ b/Event.cc
@@ -76,6 +76,7 @@ void Event::Reset(int evtID)
   reset_nan_n_silly_counters();
 
   for (auto&& l : layerHits_) { l.clear(); }
+  for (auto&& l : layerHitMasks_) { l.clear(); }
 
   simHitsInfo_.clear();
   simTrackStates_.clear();
@@ -130,6 +131,10 @@ void Event::Simulate()
     l.clear();
     l.reserve(Config::nTracks);
   }
+  for (auto&& l : layerHitMasks_) {
+    l.clear();
+    l.reserve(Config::nTracks);
+  }
 
 #ifdef TBB
   parallel_for( tbb::blocked_range<size_t>(0, Config::nTracks, 100), 
@@ -178,6 +183,7 @@ void Event::Simulate()
         // set to the correct hit index after sorting
         sim_track.addHitIdx(layerHits_[hitinfos[i].layer_].size(), hitinfos[i].layer_, 0.0f);
         layerHits_[hitinfos[i].layer_].emplace_back(hits[i]);
+        layerHitMasks_[hitinfos[i].layer_].emplace_back(0);//keep in sync, even if not used
 
         simHitsInfo_.emplace_back(hitinfos[i]);
 	if (Config::sim_val || Config::fit_val) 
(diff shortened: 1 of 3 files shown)


makortel:
Do I interpret correctly that this PR is really about making the hit masks available, and not yet about the the way/API to actually use them?

I'm just thinking that from CMSSW point of view the hit mask should be something along "a `bool` per hit".

I also reminded myself on how the hits are currently passed from CMSSW to mkFit, and it looks like by copying the `std::vector<HitVec>` (and also seeds)
https://github.com/trackreco/cmssw/blob/CMSSW_10_4_0_patch1_mkFit_X/RecoTracker/MkFit/plugins/MkFitProducer.cc#L135
https://github.com/trackreco/mkFit/blob/b6a67e624ba69c1096be415902e19bc55a3f80ce/Event.cc#L684-L688
(originally the `vector<HitVec>` on CMSSW side was not copied but moved, but that had to change when the CMSSW-side architecture went to multiple producers)

(this is likely wrong place to discuss about this but) I would actually really like if eventually the CMSSW could avoid using the `mkfit::Event` altogether.

slava77:
> Do I interpret correctly that this PR is really about making the hit masks available, and not yet about the the way/API to actually use them?

sort of

my main goal was to deliver a standalone setup where one file can be used to study different iterations (which is practical assuming that we are in reasonably perturbative regime around the reference CMSSW inputs).

still, the idea was to do it in a way that reading the bin file should allow minimal or no transformation necessary to use this mask data in individual iterations.

I guess the alternative is to save a vector of `{algo Index,  vector<vector<bool>}`, which would be even closer to the masking done in cmssw.
A somewhat simpler variant would be a `vector< vector< vector<bool`, the dimensions being (algo, layer, hit)` and I will be able to stop at 46 (max algo size) or even 37 (the largest practical iteration).

makortel:
Perhaps we should understand how the `Event::layerHits_` is going to evolve. I'm mainly wondering if during the call to `runBuildingTestPlex*()` the `layerHits_` is only read from, or if it is also modified. In case of the former, passing a pointer-to-`const` from CMSSW with the mask would make sense (as we've been planning so far), but latter case it could be simpler to reject masked-out hits already in the CMSSW side if a copy is needed anyway (which has been discussed as well, in this case it might be possible to avoid the construction of nested vectors of bools).

slava77:
>  I'm mainly wondering if during the call to runBuildingTestPlex*() the layerHits_ is only read from, or if it is also modified. 

If it's not modified, would it make sense for all different iteration instances of mkFit to share the same layerHits_? Making 10 copies (our target set of iteration) would likely become visible even at the RECO job scale.

> but latter case it could be simpler to reject masked-out hits already in the CMSSW side if a copy is needed anyway (which has been discussed as well, in this case it might be possible to avoid the construction of nested vectors of bools).

this latter option will conflict with a goal of having one .bin file for standalone testing.
It seemed like a good idea initially, at least to me, since I've been making the files.
Perhaps the cost of reorganizing the hits in `Event::read_in` is still an acceptable loss in performance though.

slava77:
> Making 10 copies (our target set of iteration) would likely become visible even at the RECO job scale.

to quantify this: 44 bytes per hit and 100K hits total => one copy is 4.4 MiB, 10 copies is 44 MiB. This is noticeable, but manageable.

makortel:
> > I'm mainly wondering if during the call to runBuildingTestPlex*() the layerHits_ is only read from, or if it is also modified.
> 
> If it's not modified, would it make sense for all different iteration instances of mkFit to share the same layerHits_? Making 10 copies (our target set of iteration) would likely become visible even at the RECO job scale.

In any case there would be a single instance of `layerHits_` in the `edm::Event`. If that gets copied to the call-to-mkFit, it would lead to 10 copies, but only one copy resident in the memory at any given time (assuming the iterations would be fully serial). But even then, I think, it would make sense for the different iteration instances of mkFit to share the same `layerHits_` (passed from CMSSW as a pointer/refererence-to-`const` for each iteration instance).

PR #285, opened 2020-11-20
Support external hit-vector indexing.

osschar:
This still works with existing code.

Example of usage is shown in EventOfHits::SuckInHits(), commented out.

Hit remapping to external / CMSSW indices is shown in MkBuilder::remap_track_hits(..), also commented out.
Some more comments from there:
  // Note that the indices after this are correct CMSSW "large-vector" indices.
  // So no hit/layer mapping code in CMSSW producer is needed.
  //
  // We could really store original indices into HitOnTrack.index
  // from the start. One just needs to be careful when getting hits in backwards fit,
  // to use (a currently non-existent) LayerOfHits::GetHitWithOriginalIndex(hot.index)
  //
  // Further, somebody should walk over quite a bit of validation code that is
  // rather involved doing such remappings.

### DIFF
--- a/mkFit/HitStructures.cc
+++ b/mkFit/HitStructures.cc
@@ -39,39 +39,45 @@ void LayerOfHits::SetupLayer(const LayerInfo &li)
 
   m_layer_info = &li;
 
-  if (is_barrel()) setup_bins(li.m_zmin, li.m_zmax, li.m_q_bin);
+  m_is_barrel = m_layer_info->is_barrel();
+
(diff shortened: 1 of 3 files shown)


makortel:
Just to note here explicitly that by visual inspection the API to add hits and extract hits of an mkfit-track look ok from CMSSW perspective. The only missing piece is the call to `runBuildingTestPlex*()` with only `EventOfHits` + `TrackVec` + hit mask, which will be looked more after @mmasciov's upcoming changes.

makortel:
Should we consider merging this PR?

PR #286, opened 2020-11-24
Remove Full Vector builder, fix TBB warnings

dan131riley:
1) Removes the Full Vector builder.  While it was an instructive exercise (we learned things from it), it ultimately was no better than the Clone Engine.  It hasn't been maintained, so it is effectively dead code now.
2) Fixes all the TBB deprecation warnings.  In most cases this just involves changing include statements, but in the top level drivers use of the global `tbb::task_scheduler_init` needed to be replaced by local task arenas (copying more or less what was done for CMSSW).
3) Fixes some miscellaneous warnings found by clang-10, fixes compilation on MacOS, and changes one `for (const auto seedIDpair : seedIDMap)` to `for (const auto& seedIDpair : seedIDMap)` (also found by clang-10).


dan131riley:
Validation in https://www.classe.cornell.edu/~dsr/mic-track/pr286/

No obvious changes from the previous PR (as expected)

Issue #290, opened 2021-01-14
Update mkFit for TBB v2021.1.1 compatibility


mrodozov:
Dear devs,

we are trying to have tbb v2021.1.1 integrated for CMS externals and currently building mkFit is failing
with:
`make: *** No rule to make target 'tbb/task_scheduler_init.h', needed by 'main.o'.  Stop.`
which I cannot say this is limited to (I mean only this is missing)
ROOT are having the same issue, so they might already have a solution (haven't checked yet)
I'd like to ask if you could update mkFit for the new tbb version.
There is this deprecation document and the substitutes for the old api:
https://software.intel.com/sites/default/files/managed/b2/d2/TBBRevamp.pdf

Thank you in advance,
Cheers,
Mircho

dan131riley:
@mrodozov I fixed the TBB deprecations on the devel branch in #286 by mostly copying what Chris did for CMSSW.  Just submitted PR #291 with the backport to the release branch.

mrodozov:
very nice :) closing the issue

Issue #240, opened 2019-08-28
Put conditions for headers depending on the architecture


mrodozov:
MkFit as it is builds only on Intel architectures, yet at CMS we also have non production ARM and PowerPC builds. To bring MkFit also into this non prod builds I introduced some hacks here:
https://github.com/cms-sw/cmsdist/pull/5174
which does two things:
1. Substitutes immintrin.h with whats available on that arch/machine 
2. Change some calls not available on non Intel archs with whatever I though was appropriate (although I'm not sure if it is)

 For the headers substitutes we proposed similar changes in other repos (see this example https://github.com/AIDASoft/DD4hep/pull/496) which looks something like :
```
#if defined(__powerpc64__) 
#include <ppc_header.h>
```
etc i.e. using preprocessor directives to include whats needed. 
Can we have something like this in the files including  immintrin.h ? 
And proper substitutes for the _mm_*intel_stuff* calls, since I'm not sure they are appropriate now.

Thank you, 
Mircho

slava77:
most of the updates in cms-sw/cmsdist#5174 are apparently related to the intel intrinsics.
There is a ```USE_INTRINSICS``` flag accessible for the makefile in command line (or environment). It seems like the problem is that not all of the intrinsics are covered by this flag.
I think that initially for the ppc and aarch we can just not use the intrinsics; although something better is always welcome.

smuzaffar:
we are now trying out a simple patch to build mkfit for non-x86_64 archs. Patch is available here https://github.com/cms-sw/cmsdist/pull/5179/files#diff-b996bc38d41188ee072f517d8691432e .
Basically we 

- build with `-sdt=c++1z`
- use `allign_alloc` (available in c++11) instead of `_mm_malloc`
- use `free` instead of `_mm_free`
- use `__buildin_prefetch` instead of `_mm_prefetch` 

These changes allowed us to build it for ARM and Power PC.

PR #291, opened 2021-01-14
Backport of tbb cleanup

dan131riley:
This PR backports the TBB cleanup to the v2.0.1_patches branch in order to address issue #290 updating TBB.  The simplest way to do the backport was to merge most of the CUDA removal PR #283 and all of the FullVector removal PR #286 so besides the TBB update it also removes a lot of dead code.

There shouldn't be an change in physics or computing performance, but I'll try running validation at Cornell (the input file for this older version is no longer in place on phi3).


dan131riley:
Benchmarks are [here](https://www.classe.cornell.edu/~dsr/mic-track/pr291/Benchmarks/).  These were run on lnx7188, so the CPU benchmarks may differ some, and the multiple events in flight show a known intermittent  anomaly on the first point.  Physics should be identical (and do appear to be) to the #240 [validation plots](https://areinsvo.web.cern.ch/areinsvo/MkFit/Benchmarks/PR240/standalone/).

slava77:
considering the amount of changes, I think that this will become v2.1.0 release once merged, instead of v2.1.0-2 patch.

slava77:
```
RecoTracker/MkFit/plugins/MkFitProducer.cc:86:41: error: no matching function for call to 'mkfit::MkBuilderWrapper::populate(bool&)'
   86 |   mkfit::MkBuilderWrapper::populate(isFV);
```
I guess it either requires a PR for CMSSW or this backport should be made more targeted to be compatible with the CMSSW release.

makortel:
> revert interface change

I suspect that is not sufficient because of
```cpp
  } else if (build == "fullVector") {
    isFV = true;
    buildFunction_ = mkfit::runBuildingTestPlexFV;
```
in https://github.com/cms-sw/cmssw/blob/eb134e644ee5255fc0656f1e4181bfca7203306e/RecoTracker/MkFit/plugins/MkFitProducer.cc#L63-L65
We still seem to declare the function
https://github.com/trackreco/mkFit/blob/27f790c3edc96630d7c658f7d4d4b86389986488/mkFit/buildtestMPlex.h#L15
but could missing the definition lead to link failure?

dan131riley:
Yes, it does lead to a link failure.  (I've now done the test build with CMSSW that I should have before making the PR.)

dan131riley:
I've verified that the CMSSW MkFit plugin compiles and links with the latest two commits, apologies for not testing that before making the PR.  I could make the backport more targeted, but it will take more time.  (That we're still declaring `runBuildingTestPlexFV` in devel is a bug...)

makortel:
Just to say it out loud, I'm thinking to remove the use of FullVector from CMSSW side quickly after the mkFit external gets updated (and update https://github.com/trackreco/cmssw as well).

PR #294, opened 2021-01-31
Multi iter

osschar:
I propose we just merge this into devel now to minimize forking.

I've added the tasks to "Deployment/development plans for mkFit in Run-3" doc.


osschar:
These are the validation plots (copied from Leonardo):
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/294-MultiIter/

PR #299, opened 2021-02-23
Progress with validation on the multiple iterations

leonardogiannini:
this branch is still work in progress.

I should look into simplifying the validation as a next step

When should I merge it?

### DIFF
--- a/Event.h
+++ b/Event.h
@@ -4,7 +4,7 @@
 #include "Track.h"
 #include "Validation.h"
 #include "Config.h"
-
+#include "mkFit/SteeringParams.h"
 #include <mutex>
 
 namespace mkfit {
@@ -36,6 +36,7 @@ public:
   int  use_seeds_from_cmsswtracks(); //special mode --> use only seeds which generated cmssw reco track
   int  clean_cms_simtracks();
   int  clean_cms_seedtracks(TrackVec *seed_ptr = nullptr); //operates on seedTracks_; returns the number of cleaned seeds
+  int  clean_cms_seedtracks_iter(TrackVec *seed_ptr, IterationConfig itcfg);
   int  clean_cms_seedtracks_badlabel(); //operates on seedTracks_, removes those with label == -1;
   void relabel_bad_seedtracks();
   void relabel_cmsswtracks_from_seeds();
--- a/Event.h
+++ b/Event.h
@@ -36,12 +36,14 @@ public:
   int  use_seeds_from_cmsswtracks(); //special mode --> use only seeds which generated cmssw reco track
   int  clean_cms_simtracks();
   int  clean_cms_seedtracks(TrackVec *seed_ptr = nullptr); //operates on seedTracks_; returns the number of cleaned seeds
-  int  clean_cms_seedtracks_iter(TrackVec *seed_ptr, IterationConfig itcfg);
+  int  clean_cms_seedtracks_iter(TrackVec *seed_ptr, const IterationConfig& itrcfg);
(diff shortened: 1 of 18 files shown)


osschar:
your call, we can merge it now. I can resolve the conflicts if you want.

do you want me to check something in particular?

TTreeValidation.cc
```diff
@@ -2318,9 +2360,9 @@ void TTreeValidation::fillConfigTree()
   fPhiFactor_ = Config::fPhiFactor;
   nEtaPart_   = Config::nEtaPart;
 
-  nlayers_per_seed_ = Config::nlayers_per_seed;
-  maxCand_ = Config::maxCandsPerSeed;
-  chi2Cut_ = Config::chi2Cut;
+  nlayers_per_seed_ = Config::ItrInfo[0].m_params.nlayers_per_seed;
+  maxCand_ = Config::ItrInfo[0].m_params.maxCandsPerSeed;
+  chi2Cut_ = Config::ItrInfo[0].m_params.chi2Cut;
```
osschar, line 2370: Should this get iteration id from somewhere, instead of hardcoding to 0?

Event.h
```diff
@@ -36,11 +36,14 @@ class Event
   int  use_seeds_from_cmsswtracks(); //special mode --> use only seeds which generated cmssw reco track
   int  clean_cms_simtracks();
   int  clean_cms_seedtracks(TrackVec *seed_ptr = nullptr); //operates on seedTracks_; returns the number of cleaned seeds
+  int  clean_cms_seedtracks_iter(TrackVec *seed_ptr, const IterationConfig& itrcfg);
```
osschar, line 39: Eventually we should move this out of Event (and probably remove the old seed_clean function and use the new one in the old iteration0 functions).
But we can do it in the next step, after we think through duplicate cleaning and params there.

mkFit/buildtestMPlex.cc
```diff
@@ -353,10 +353,19 @@ double runBuildingTestPlexCloneEngine(Event& ev, EventOfHits &eoh, MkBuilder& bu
 double runBtbCe_MultiIter(Event& ev, EventOfHits &eoh, MkBuilder& builder)
 {
   double ttime = 0;
+  TrackVec seeds_used;
+  TrackVec seeds1;
 
-  // ??? MIMI - What validation prep / map tasks need to run before we start
-  // screwing up seeds etc?
-
+  //not smart usage of memory
```
osschar: We need to figure out a way to say "if validation" to avoid doing remapping and seed duplication when we don't need it.
It seems const bool validation_on = (Config::sim_val || Config::quality_val); would be a good start?

leonardogiannini: adding the flag validation_on

TTreeValidation.cc
leonardogiannini: this is the configtree which is never used later in validation

I think we can fill it once per iteration in case we run multi-iteration setup

Event.h
leonardogiannini: I prefer doing it as a next step

leonardogiannini:
I also wnated to check the validation before and after the PR - 

http://uaf-7.t2.ucsd.edu/~lgiannini/before_PR//
http://uaf-7.t2.ucsd.edu/~lgiannini/rerun_init_afterPR/

it is ok, 
so I would merge and then decide what to do with moving the seed cleaning function and the with the configTree, which is actually not used anywhere in the validation

PR #302, opened 2021-03-05
New tuning after pandoras box

areinsvo:
This is the new tuning for the candidate scoring after all of Pandora's Box, as described in slides presented on [February 19](https://indico.cern.ch/event/971170/contributions/4088757/attachments/2193594/3708015/tuningFeb2021.pdf) and [March 5](https://indico.cern.ch/event/971172/contributions/4088787/attachments/2202859/3726460/tuningMarch2021.pdf) and in the google doc [here](https://docs.google.com/spreadsheets/d/1nZJxZ4qMrmotxAenZhlcRQmvKVcVsm-7IV_n_MbvkKo/edit#gid=1882976983). 

The stable 2020 performance is [here](https://areinsvo.web.cern.ch/areinsvo/MkFit/PandorasBox/stable2020_benchmarks/).

The Pandora's Box performance (pretuning) is [here](https://areinsvo.web.cern.ch/areinsvo/MkFit/PandorasBox/PandorasBox_benchmarks/).

The new tuning results in this PR are [here](https://areinsvo.web.cern.ch/areinsvo/MkFit/PandorasBox/benchmarks_tune2_chi30/).

### DIFF
--- a/Config.h
+++ b/Config.h
@@ -350,9 +350,11 @@ namespace Config
   constexpr float track_ptlow = 0.9;
 
   // sorting config (bonus,penalty)
-  constexpr float validHitBonus_ = 10;
+  constexpr float validHitBonus_ = 4;
+  constexpr float validHitSlope_ = 0.2;
   constexpr float overlapHitBonus_ = 0; // set to negative for penalty
-  constexpr float missingHitPenalty_ = 10;
+  constexpr float missingHitPenalty_ = 8;
+  constexpr float tailMissingHitPenalty_ = 3;
 
   // Threading
   extern int    numThreadsFinder;
--- a/Track.h
+++ b/Track.h
@@ -32,6 +32,7 @@ public:
   int   hitIdx; // hit index
   unsigned int   module; // module id
   int   nhits;  // number of hits (used for sorting)
+  int   ntailholes; // number of holes at the end of the track (used for sorting)
   int   noverlaps; // number of overlaps (used for sorting)
   int   nholes;  // number of holes (used for sorting)
   unsigned int seedtype; // seed type idx (used for sorting: 0 = not set; 1 = high pT central seeds; 2 = low pT endcap seeds; 3 = all other seeds)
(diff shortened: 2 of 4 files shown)


areinsvo:
This is ready to be merged if others agree. The MTV results with this tuning were discussed at [Friday's meeting](https://indico.cern.ch/event/971173/contributions/4088790/attachments/2207582/3735444/MTV_PostTuning.pdf). The full set of MTV plots can be found [here](https://areinsvo.web.cern.ch/areinsvo/MkFit/CMSSW_Integration/PandorasBox/plots_tuning).

mmasciov:
Just a minor comment:
The choice here was to remove the scoring based on seed type and instead use the candidate pT directly.
If I recall correctly, the reason for using seed type rather than candidate kinematics was due to the fact that the candidates are updated at different times depending on the build, and also so that the bonus/penalty system would be based on the starting pT/pseudo-rapidity of the seed rather than on each candidate.
I'm not opposed to switching to candidate parameters rather than seed ones.
However, seed type also accounted for the pseudo-rapidity of the track, while in the current implementation we lost this possibility.
This is not always optimal; e.g., compare fake rate for low pT tracks at |eta|>0.9:
- before: https://areinsvo.web.cern.ch/areinsvo/MkFit/PandorasBox/PandorasBox_benchmarks/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_fr_eta_build_pt0p0_SIMVALSEED.png
- after: https://areinsvo.web.cern.ch/areinsvo/MkFit/PandorasBox/benchmarks_tune2_chi30/SIMVAL_MTV_SEED/SKL-SP_CMSSW_TTbar_PU50_fr_eta_build_pt0p0_SIMVALSEED.png

I'm not suggesting that we go for a better optimization now, but it may be good to add "eta" to the list of the current arguments of the scoring functions, to allow for further future optimization, and not to lose information that we used to exploit?

mmasciov:
After talking to @areinsvo, I believe this PR can be merged as is.
If necessary, we'll revisit it at a later stage. Hence, I'm merging it.

PR #305, opened 2021-03-15
PAtch C++ configuration from a JSON patch file

osschar:
New command line arguments:
- --json-patch 'filename'  patch iteration config from given JSON file
- --json-patch-dump-before print iteration config before patching
- --json-patch-dump-after  print iteration config after  patching

JSON file format notes:
- See example JSON patcher input: "mkFit/config-parse/test.json"
- The file can contain several valid JSON dumps in sequence.
- '/' character can be used to descend more than one level at a time.
- A number can be used to specify an array index. This can be combined with the '/' syntax.
- "[first,last]" key (as string) can be used to denote a range of array elements. Such a key must not be combined with a '/' syntax.


makortel:
I'm trying to think a possible workflow (on either standalone or CMSSW). Would it be along having  the "authoritative configuration" per iteration would be in `CMSSW-2017.(a)cc`, and patching that configuration with the JSON file? So the JSON file would be used only as an aid to avoid recompilation when doing parameter scans, and the outcome of that scan would then be set in the C++ as the new set of defaults?

(I think such approach could make sense in the near term; eventual CMSSW deployment might need something different)

osschar:
Yes, I think that was the plan for "step 1", until we stabilize the multiple-iterations and tuning of seed cleaning, finding and duplicate removal.

One could already now provide a method for loading the full config from json ... but I don't think it's either pretty or very useful. A pretty-printed dump of one iteration is about 800 lines (mostly coming from 72 layers).

What is sort of bothering me is mkFit layer definitions and extraction of mkFit geometry from Reco geometry. The tricky thing here are holes in strip stereo coverage in endcaps.

PR #306, opened 2021-03-19
Use original hit indices in mkFit hits and tracks.

osschar:
Mapping of seed hits and re-mapping of result tracks is not needed.

Note, mapping/remapping functions are still in the code (for standalone and cmssw). The calls to these functions have been commented out with // ZOIX prefix. I could in principle remove those functions and calls as I don't think we will be going back to copying of hits into internal vectors.

Results are identical on a limited comparison with quality validation.

If somebody whose access to phi3 has been re-established could run the validation it would be appreciated.

### DIFF
--- a/mkFit/HitStructures.cc
+++ b/mkFit/HitStructures.cc
@@ -276,7 +276,8 @@ void LayerOfHits::EndRegistrationOfHits(bool build_original_to_internal_map)
 
     m_phi_bin_infos[q_bin][phi_bin].second++;
 
-    // m_hit_ranks[i] will never be used again ... use it to point to external index.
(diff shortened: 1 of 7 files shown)


mmasciov:
I have run the standard benchmarks:
- before: http://uaf-1.t2.ucsd.edu/~mmasciov/MkFitBenchmarks/beforePR306/
- after: http://uaf-1.t2.ucsd.edu/~mmasciov/MkFitBenchmarks/afterPR306/
I didn't spot any (significant) difference myself.

osschar:
If somebody has a preference for keeping the mapping/remapping code in let me know by tomorrow morning Pacific ... otherwise I'll remove this dead code and merge the PR.

PR #313, opened 2021-04-30
readme.md updates: AVX2, CMSSW testing (gcc and task pinning)

slava77:
- AVX2 by default in the instructions
- gcc SSE3 build of mkfit
    - I did this by `unset INTEL_LICENSE_FILE`, which may be a pain; a better solution is welcome
- taskset for pinning cmsRun job

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,6 +503,9 @@ git cms-init
 popd
 git clone git@github.com:trackreco/mkFit
 pushd mkFit
+# for gcc CMSSW "default" build:
+#   1) call "unset INTEL_LICENSE_FILE", or do not source compilevars.sh above
+#   2) replace AVX* with VEC_GCC="-msse3"
 make -j 12 TBB_PREFIX=$(dirname $(cd $CMSSW_BASE && scram tool tag tbb INCLUDE)) WITH_ROOT=1 AVX2:=1
 popd
 ```
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ You are free to put the lines from this script in your login scripts (.bashrc, .
 Now compile the code:
 
 ```
-make -j 32 AVX_512:=1
+make -j 32 AVX2:=1
 ```
 
 To run the code with some generic options, do:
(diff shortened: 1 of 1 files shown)


slava77:
@makortel @osschar 
please take a look

makortel:
Looks good to me

osschar:
Maybe we (Steve and I) should also add support for SSE3 intrinsics for Matriplex :)

It's funny how this is all going ... our first attempts were with MIC instruction set (precursor to AVX512) and now we are going down towards 128 bits. This, however, does make some sense as ARM, IIRC, is planning to only support this width although with their native instruction set.

srlantz:
Sorry I didn't get around to checking this before it was merged. There's a trivial typo: it should say compilervars.sh, not compilevars.sh. That's not too important, but here's something that is: the instructions on how to set up the Intel compilers are arguably out of date, now that Intel oneAPI is available on phi3. The icc in oneAPI is more recent than the others on phi3. Note, this icc is still the "real" Intel C/C++ compiler, in the same lineage as previous versions of icc. (I guess icc now becomes a backronym for Intel Compiler Classic.) If we want to standardize on the newer icc (just as we already have on lnx7188), then one of the following should be used to source the necessary Intel script, which is no longer called compilervars.sh:
```
source /opt/intel/oneapi/compiler/latest/env/vars.sh     # makes only compiler-related changes to the env
source /opt/intel/oneapi/setvars.sh    # adds lots of Intel stuff to the env, including oneTBB (the latest TBB)
```
I should also note that if we want to make phi3 consistent with lnx7188, then the existing Intel oneAPI should be uninstalled and replaced with the latest oneAPI. This will update icc from 2021.1.2 to 2021.2.0. Dan just did this on lnx7188.

(I'll respond to Matevz's comment in a bit)

srlantz:
@osschar  I doubt that we will ever prefer SSE3 to AVX2 for mkFit, especially when it's compiled with icc. I have to believe that it makes a difference in performance to have an FMA instruction available, which SSE3 lacks. (There is an FMA instruction for the xmm registers, but it was added in AVX-512, so it's not really part of SSE.) The CPU may be able to pipeline an SSE-multiply followed by an SSE-add so that it works just like an FMA, but I wouldn't count on that.

From what I saw in playing around with mkFit on AMD Rome processors, intrinsics are becoming less and less important to mkFit performance. The code actually ran faster with our intrinsics disabled  :-{    I guess the compilers are getting quite good at optimizing plain old loops. They are also getting sneaky about choosing xmm or ymm or zmm registers depending on what's in the loop. We may be reaching the point where it's just better for us to stay out of the way of the compilers.

Anyway, this is why I am recommending that we add some new benchmarks where we compile the code a few different ways using different compilers, then run them with a significant number of threads and events in flight, making sure that Turbo is enabled on the test machine. This will help us keep an eye on what the real-world performance is like, as well as which method of compiling works the best.

IHateLinus:
IMO we can do all that after the PR agains CMSSW is in.

Lets please focus on that.
Much larger issues are e.g. making an apples to apples comparison w ckf in terms of machine setting.

osschar:
Yes, that is clear, no worries :)

srlantz:
Yes, but at the same time we want to have a clear, reproducible demonstration that mkFit offers performance advantages. I won't burden this comment thread further with a discussion, though.

PR #314, opened 2021-05-06
Duplicate removal, quality filter and hit selection windows improvements

osschar:


mkFit/HitStructures.h
```diff
@@ -494,7 +494,36 @@ class CombCandidate : public std::vector<TrackCand>
...
+  }
+
+  // Need this for std::swap when filtering EventOfCombinedCandidates::m_candidates.
+  // We do not call clear() on vectors as this will be done via EoCCs reset.
+  // Probably would be better (clearer) if there was a special function that does
+  // the swap in here or in EoCCs.
+  CombCandidate& operator=(CombCandidate&& o)
+  {
+    std::vector<TrackCand>::operator=( std::move(o) );
+    m_best_short_cand = std::move(o.m_best_short_cand);
```
makortel, line 512: Strictly speaking this is use-after-move.

makortel: Does `CombCandidate` really need to inherit from `std::vector<TrackCand>`? (usually inheriting from std containers just causes problems)

osschar: Will remove the op=(&&) and put in explicit swap() method.

Will also try to get rid of the inheritance, it's a case of historical laziness :)

osschar: In another PR.

makortel: > In another PR.

Sure.

mmasciov:
For history:
- MTV results for 9 iterations: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_multiIteration_forPR/plots_500evts_all/
- MTV results for 7 iterations (excluding mixelTripletStep and detachedTripletStep): http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_multiIteration_forPR/plots_500evts_lessIters_all/
The latter is what is proposed for PR to CMSSW.

PR #315, opened 2021-05-08
Implement saving of configuration to a set of JSON files, one per iteration.

osschar:
To export (%d MUST be present):
  `` ./mkFit --json-save-iterations mkfit-iteration-%d.json ``

To load a single file, e.g.:
  `` ./mkFit .... --json-patch mkfit-iteration-0-test.json ``

To load a set of files use shell, --json-patch must be given for every file:
  `` ./mkFit .... `for a in {0..8}; do echo --json-patch mkfit-iteration-$a.json; done` ``

From ./mkFit --help:
```
JSON config patcher options:
  --json-patch <filename>  patch iteration config from given JSON file (def: do not patch)
                           can be specified multiple times for several files
  --json-save-iterations <fname-fmt> save per iteration json files
                          %d in fname-fmt gets replaced with iteration number
  --json-patch-verbose     print each patch assignment as it is being made (def: false)
  --json-patch-dump-before print iteration config before patching (def: false)
  --json-patch-dump-after  print iteration config after  patching (def: false)
```

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -102,7 +102,8 @@ namespace Config
   bool        json_patch_dump_before = false;
   bool        json_patch_dump_after  = false;
   bool        json_patch_verbose     = false;
-  std::string json_patch_filename;
+  std::vector<std::string> json_patch_filenames;
+  std::string              json_save_iters_fname_fmt;
 
   void RecalculateDependentConstants()
   {
--- a/Config.cc
+++ b/Config.cc
@@ -99,11 +99,13 @@ namespace Config
   bool  backwardFit = false;
   bool  includePCA = false;
 
-  bool        json_patch_dump_before = false;
-  bool        json_patch_dump_after  = false;
-  bool        json_patch_verbose     = false;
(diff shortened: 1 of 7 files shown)


osschar:
I propose to copy CMS-2017-HitSelectionWindows.h  CMS-2017.acc  CMS-2017.cc from Geoms/ to RecoTracker/MkFit/, build them there, and call the geometry/iteration-config generation function from geometry-producer directly, without using the geometry-plugin loading mechanism.

In the future we then cleanup this code so IterationConfigs can be created independently, per iteration.

makortel:
The plain `./mkFit --json-save-iterations mkfit-iteration-%d.json` did not work for me, adding `--geom CMS-2017` made it to dump the JSONs. The execution itself either gets stuck or crashes, but I suppose the JSON files themselves are fine.

osschar:
I tested saving and loading while running "standard" jobs over 1 event. So yes, I assume they are ok. :)

Issue #266, opened 2020-03-04
clean out KNC cruft from make


srlantz:
Makefile and Makefile.config still have various sections related to building mkFit for the now-defunct Knights Corner coprocessor, which was removed from the phiphi test machine a long time ago (when it became phi1). At this point, we would be better off without the clutter.

srlantz:
Closed by #316.

PR #316, opened 2021-05-09
Remove USE_MATRIPLEX and remaining traces of MIC as C define and make variable

osschar:
Add SSE3 option to Makefile.config.

Set Matriplex width to 4 for SSE.

gcc-10, AMD Ryzen 7, cand build time, 10 events (after skipping 1st), best-out-of 2

```
   sse4   sse3   avx   avx2    avx-int  avx2-int
 4 2.30   2.34
 8 2.64          2.54  2.47    2.56     2.54
16 3.08                2.93
32                     3.71
```

### DIFF
--- a/Config.cc
+++ b/Config.cc
@@ -23,7 +23,7 @@ namespace Config
   int   numThreadsFinder = 1;
   int   numThreadsEvents = 1;
   
-#if defined(__MIC__) || defined(__AVX512F__)
+#if defined(__AVX512F__)
   int   numThreadsSimulation = 60;
 #else
   int   numThreadsSimulation = 12;
--- a/Config.h
+++ b/Config.h
@@ -413,22 +413,21 @@ namespace Config
     return (Config::mag_b0*z*z + Config::mag_b1*z + Config::mag_c1)*(Config::mag_a*r*r + 1.f);
   }
 
-#ifdef USE_MATRIPLEX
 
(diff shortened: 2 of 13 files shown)


srlantz (approved):
Thanks @osschar, this PR does a lot of good cleanup and actually closes issue #266 for us. However, I will likely open a new issue to track some of the following requests, most of which pertain to Makefile.config.

Question, in config.h, what is the purpose of THREAD_BINDING?

In Makefile.config, I think we need to get rid of the test for INTEL_LICENSE_FILE and pick some other test for icc. There is no longer going to be a license file for those who download Intel oneAPI to get the free icc = Intel C/C++ Compiler Classic. On a related note, the init_env.sh script in xeon_scripts should be updated for the icc in Intel oneAPI, gcc 9 or 10, etc.

Also, the icc man page claims -mfma is a no-op. I believe the preferred AVX2 option should be -march=core-avx2. This works for both icc and gcc (and clang too, ultimately).

Other updates to Makefile.config can be considered, though they're not urgent, as the lines in question are commented out. Anyway, some of the "Mac OS X" (now macOS) choices are way out of date. Likewise devtoolset-2, etc.

I'm not sure what effect USE_VTUNE_NOTIFY is having these days, along with the ittnotify library. The calls to that lib in the code should be revisited at some point.

PR #318, opened 2021-06-07
revert writing of dummy hits from PR 287

slava77:
@cerati 
informed me that the code added to WriteMemoryFile.cc in #287 is not needed and the dead modules will be handled by other means


cerati:
yes, we are currently exploring a different strategy to account for dead modules. Merging.

PR #321, opened 2021-06-14
Merge duplicate seeds

osschar:
Merge hits from cleaned duplicate seeds into the remaining seed.

This runs for --build-mimi and for CMSSW. It is not used for the legacy --build-bh/std/ce.

Statistics of merged seed hits: new stand alone files PU 50, 200 events:
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/SeedHitMerge/

Open question: Do we want to make this optional / add a configuration variable?

### DIFF
--- a/Track.cc
+++ b/Track.cc
@@ -132,7 +132,7 @@ void Track::resizeHitsForInput()
 
 void Track::sortHitsByLayer()
 {
-  std::sort(& hitsOnTrk_[0], & hitsOnTrk_[lastHitIdx_ + 1],
+  std::stable_sort(& hitsOnTrk_[0], & hitsOnTrk_[lastHitIdx_ + 1],
 	    [](const auto & h1, const auto & h2) { return h1.layer < h2.layer; });
 }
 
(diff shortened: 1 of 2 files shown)


areinsvo:
MTV plots here: http://uaf-10.t2.ucsd.edu/~lgiannini/plots_compareDupMergeSeeds_7iters
Standalone validation plots (without Matevz's fix of sorting the seeds, but since these results are all for build tracks, it shouldn't matter): https://areinsvo.web.cern.ch/areinsvo/MkFit/MultiIter/mergedDuplicates/
(Compare to the baseline [here](https://areinsvo.web.cern.ch/areinsvo/MkFit/MultiIter/master-multi_v2)

I guess if we had a config variable for this, then it could be turned off for the strip-seeded variables? Otherwise, I don't see a need to make this optional. 

Do we want to check timing before merging? I didn't run standalone benchmarks, but the MTV plots from @leonardogiannini have timing info (assuming all of those results were run on a mostly empty machine). The [summary timing plot](http://uaf-10.t2.ucsd.edu/~lgiannini/plots_compareDupMergeSeeds_7iters/plots_timing/summaryReal.pdf) shows no change in build time but an increase in fit time.

osschar:
Seed cleaning is not run for all iterations ... when IterationConfig::m_require_quality_filter is true, it means we clean duplicates afterwards and do NOT do seed cleaning.

I wouldn't expect a significant impact on timing from this change.

PR #327, opened 2021-06-30
FPE hunt side fixes

osschar:
1. Fix IterationParams::c_dzmax_bh = 0 for iteration_index 3 / track_algo 5
2. Use - to do charge sign flip, do not multiply with -1.
3. When merging seed hits, always sort ... the iteration-specific sorting happens before, not after seed cleaning.
4. Handle more than 2 overlap hits per layer in bk-fit.
5. Document --num-iters-cmssw in --help text.

Points 3 and 4 could change results a bit as they affect backward-fit's propagation into seeding region ... can somebody please run relevant validations?


mmasciov:
This PR is effectively replaced by PR #332. Hence, closing.

PR #334, opened 2021-07-22
invpt sign fix and numerical precision improvements for high pt

slava77:
http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10muHS_CKF_vs_mkFit_310_absPt_77a7f37/plots_initialStep.html
has plots for the high-pt muon sample

initial step built tracks: recently merged v3.1.0 (red); fix to the invpt charge flip (black) and this PR as of 77a7f37 (orange)
image
image

It looks like some of the inner hits are recovered (I think that this is from the recovered tracks that make it all the way inside without failing to a bad state in the backward fit)
image

The pulls are better
image

phi resolution look better
image
.. however, something odd is happening in the same vs pt (it looks like the fits did not converge)
image

The higher eta region apparently still needs more work.

For ttbar pu50 there are no significantly visible differences roughly as expected since most of the changes were targeting high-pt
http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar_CKF_vs_mkFit_310_77a7f37/
e.g. all tracks OOB (this PR line is in black here)
image

timing with AVX2 looks consistent between two mkfit versions (v3.1.0 and this PR)
image
pixel-less has some more noticeable increase in timing, but I think that it's within uncertainties (also, there are no significant changes in the variables in this iteration)

vectorization report is suggestive of an overall moderate improvement in PropagationMPlex (`helixAtZ` and `applyMaterialEffects`) https://github.com/trackreco/mkFit/pull/334#pullrequestreview-712892340

A couple of points that were observed during debugging of a few cases:
- getHitSelDynamicWindows is too tight in the endcap stereo layers
- backward fit should better start from the last valid hit (instead of occasional far-off detours at the edge of the strip tracker acceptance)
- dr in collect hit indices is missing sqrt

### DIFF
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -1309,6 +1309,17 @@ void MkFinder::BkFitFitTracksBH(const EventOfHits   & eventofhits,
                             Err[iP], Par[iP], msErr, msPar, Err[iC], Par[iC], tmp_chi2, N_proc);
     }
 
+    //fixup invpt sign and charge
+    for (int n = 0; n < N_proc; ++n)
+    {
+      if (Par[iC].At(n,3,0) < 0)
+      {
+        Chg.At(n, 0, 0)  = -Chg.At(n, 0, 0);
+        Par[iC].At(n,3,0) = -Par[iC].At(n,3,0);
+      }
+    }
+
+
 #ifdef DEBUG_BACKWARD_FIT_BH
     // Dump per hit chi2
     for (int i = 0; i < N_proc; ++i)
--- a/mkFit/PropagationMPlex.cc
+++ b/mkFit/PropagationMPlex.cc
@@ -617,6 +617,7 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
       const float theta = inPar.ConstAt(n, 5, 0);
 
       const float k = inChg.ConstAt(n, 0, 0) * 100.f / (-Config::sol*(pflags.use_param_b_field?Config::BfieldFromZR(zin,hipo(inPar.ConstAt(n,0,0),inPar.ConstAt(n,1,0))):Config::Bfield));
+      const float kinv = 1.f/k;
 
       dprint_np(n, std::endl << "input parameters"
             << " inPar.ConstAt(n, 0, 0)=" << std::setprecision(9) << inPar.ConstAt(n, 0, 0)
@@ -633,6 +634,7 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
       //no trig approx here, phi can be large
       const float cosP = std::cos(phiin), sinP = std::sin(phiin);
       const float cosT = std::cos(theta), sinT = std::sin(theta);
+      const float tanT = sinT/cosT;
       const float pxin = cosP*pt;
       const float pyin = sinP*pt;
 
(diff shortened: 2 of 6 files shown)


cerati:
nice! for curiosity: it looks like you implemented a few fixes... any idea of their relative importance?

mkFit/MkFinder.cc
```diff
@@ -1309,6 +1309,17 @@ void MkFinder::BkFitFitTracksBH(const EventOfHits   & eventofhits,
                             Err[iP], Par[iP], msErr, msPar, Err[iC], Par[iC], tmp_chi2, N_proc);
     }
 
+    //fixup invpt sign and charge
```
osschar, line 1312: Shouldn't we put this in KalmanOperation[Endcap], if (update-requested)?
At least it should also go in BkFitFitTracks() as this is the one I'm switching to for backward search --- I'd have to remember to also put it in there.

slava77: I couldn't quickly figure out how to pass a non-const charge

slava77:
> nice! for curiosity: it looks like you implemented a few fixes... any idea of their relative importance?

-  dc7999b is red vs black
- 9162c92 the material effects truncation in the muon sample is likely only for almost failed candidates.
- I think that  451acee  (propagation numerics) contributes more than 77a7f37 (covariance numerics/refactoring), I don't have an individual diff
    -   451acee  (propagation numerics) contributes also on the way forward and should produce longer tracks (IIRC I actually first picked it up for a barrel track with a seed pt around 1e5 GeV and the propagation accumulates large enough offset from the pixel to the TIB that the update messes up the track)

slava77:
the counts of "candidate ignored", based on the count of failing the light pos-definiteness check (diag elements positive) have some noticeable reduction:
- ttbar 100 events: orig has 8269 -> this PR ( 77a7f37 ) has 4725
- 10mu 10K events: orig 5277 -> this PR ( 77a7f37 ) 2921

mkFit/PropagationMPlex.cc
```diff
@@ -295,7 +295,7 @@ void helixAtRFromIterativeCCSFullJac(const MPlexLV& inPar, const MPlexQI& inChg,
       errorPropTmp(n,4,4) = 1.f;
       errorPropTmp(n,5,5) = 1.f;
```
slava77, line 297: in this loop
from optreport original ==> this pr
```
   remark #15475: --- begin vector cost summary ---
   remark #15476: scalar cost: 1190           ==> 1201
   remark #15477: vector cost: 427.250     ==> 430.75
   remark #15478: estimated potential speedup: 2.420   ==> 2.420
```

mkFit/PropagationMPlex.cc
```diff
@@ -627,10 +630,11 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
```
slava77, line 630: in this loop 
```
   remark #15476: scalar cost: 1375                 ==> 1124
   remark #15477: vector cost: 144.370           ==> 121.870
   remark #15478: estimated potential speedup: 8.470 ==> 8.720
   remark #15482: vectorized math library calls: 3
   remark #15486: divides: 16                           ==> 9
   remark #15487: type converts: 1
```
is apparently in a good direction

mkFit/PropagationMPlex.cc
```diff
@@ -717,7 +729,7 @@ void applyMaterialEffects(const MPlexQF &hitsRl, const MPlexQF& hitsXi, const MP
       const float beta2 = p2/(p2+mpi2);
```
slava77, line 729: in this loop
```
   remark #15476: scalar cost: 748                                    ==> 748 
   remark #15477: vector cost: 154.120                             ==> 141.000
   remark #15478: estimated potential speedup: 4.590   ==> 5.000
   remark #15482: vectorized math library calls: 5           
   remark #15486: divides: 14
   remark #15487: type converts: 3                                   ==> not mentioned
```
apparently `1. -> 1.f` mattered

slava77 (commented):
here are some notes from PropagationMPlex.optrpt.
It looks like in a combination, there is some improvement.

slava77:
I added some notes from ttbar MTV in the PR description.

The vectorization report analysis in https://github.com/trackreco/mkFit/pull/334#pullrequestreview-712892340 looks good as well.

PR #336, opened 2021-07-26
In bkfit, copy out final parameters from propagated or updated state based on the state of propagate-to-PCA.

osschar:


osschar:
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/336-bkfit-copy-out/

mmasciov:
MTV results in TTbar + PU50: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_PR336/
Differences between head and PR #336 appear to be marginal.

slava77:
>  Differences between head and PR #336 appear to be marginal.

I was not expecting a significant effect on average; this is more about getting the logic correct.

mmasciov:
If there are no objections, I'm going to click the 'merge' button on this PR.

PR #339, opened 2021-08-10
update dead module treatment to a more restrictive

slava77:
- restrict addition of a gap/dead hit only if the candidate did not have other hits added
- q bin and phi bin updates from https://github.com/trackreco/mkFit/compare/deadmod-reduce-region
    - the q bin part is taken as is
    - the phi binning is updated just to 256

the validation and  choice for 256 phi bins is based on 
[mkFit_deadMod_080921.pdf](https://github.com/cms-sw/cms-sw.github.io/files/6962159/mkFit_deadMod_080921.pdf)
links to mtv plots:

- restriction based on hit count and qbin ("drFix" is the baseline without dead modules as of #335)
    - 10mu http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit_dr_dead-restrict_me082692_c19205b8
    - ttbar http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_mkFit_dr_dead-restrict_me082692_c19205b8
- bin variations
    - 10mu http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit_dr_dead-restricted-Qbin-p512_m7405a71_c19205b8
    - ttbar http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_mkFit_dr_dead-restricted-Qbin-p512_m7405a71_c19205b8

@leonardogiannini please check how the failure scenarios look like with this update.

### DIFF
--- a/Config.h
+++ b/Config.h
@@ -165,7 +165,7 @@ namespace Config
   extern float XigridME[Config::nBinsZME][Config::nBinsRME];
 
   // This will become layer dependent (in bits). To be consistent with min_dphi.
-  static constexpr int m_nphi = 128;
+  static constexpr int m_nphi = 256;
 
   // config on Event
   /*MM: chi2Cut is now set in IterationsParams*/
--- a/mkFit/HitStructures.h
+++ b/mkFit/HitStructures.h
@@ -150,8 +150,8 @@ public:
 
   // Testing bin filling
   static constexpr float m_fphi     = Config::m_nphi / Config::TwoPI;
-  static constexpr int   m_phi_mask = 0x7f;
-  static constexpr int   m_phi_bits = 7;
+  static constexpr int   m_phi_mask = 0xff;
+  static constexpr int   m_phi_bits = 8;
   static constexpr float m_fphi_fine      = 1024 / Config::TwoPI;
   static constexpr int   m_phi_mask_fine  = 0x3ff;
   static constexpr int   m_phi_bits_fine  = 10; //can't be more than 16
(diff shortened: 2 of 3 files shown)


osschar:
Have you tried skipping addition of an invalid hit only if best chi2 of a hit is below certain limit, say, 0.5 * max_chi^2?

slava77:
> Have you tried skipping addition of an invalid hit only if best chi2 of a hit is below certain limit, say, 0.5 * max_chi^2?

no. I somehow convinced myself that this will go in the opposite direction of where the target performance should be, that this or similar selection can only lead to more dead/gap hits added and will further increase the fakes.

cerati:
keep in mind that the number of invalid hits we currently allow is tuned without the dead module feature, so it needs to be reduced after this feature (otherwise an increase in fake rate is probably expected). I'd be happy to try some more tuning once the approach is consolidated both for the dead modules and the search windows

slava77:
> keep in mind that the number of invalid hits we currently allow is tuned without the dead module feature, so it needs to be reduced after this feature (otherwise an increase in fake rate is probably expected). I'd be happy to try some more tuning once the approach is consolidated both for the dead modules and the search windows

I tried to make a step in that direction by reducing the count of consecutive holes allowed in https://github.com/trackreco/mkFit/commit/101298b39d1331a40411cd3855fff9afde718218
but it looks like in the transition/endcap region we begin to lose a lot of hits
http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit_dr_dead-restricted-Qbin-p256-consec_m101298b_c19205b8/plots_initialStep.html

cerati:
I wonder if there is an interplay with tight windows...

PR #341, opened 2021-08-11
remove requirement for pixel hits for seedSimIdx (seed label) in the standalone bin file writer

slava77:


slava77:
I'll try to combine this with the update in the scripts to the new bin files

slava77:
this is now in #342

PR #345, opened 2021-08-18
Backward search implementation

osschar:
1. Mechanics of backward-search + configs for pixeless and tobtec iterations.
2. Flexible SteeringParams::iterator that gives more control over layer iteration.

Validation / MTV has been run by Leonardo two weeks back.

CMSSW json files will have to be remade as there is a new member in IterationParams.

At some point we should revisit what propagate-to-PCA does(and when):
- cmssw is not using it;
- in standalone it has to be turned on but it happens before backward-search for the two iterations where backwards-search is actually used.


leonardogiannini:
what about the changes in RecoTracker/MkFit/plugins/createPhase1TrackerGeometry.cc ? are they still necessary for the PR to work?

osschar:
oh yes, the same ... there is extra stuff in there for bkw-search params since the previous commit

slava77:
do we have validation plots for this PR?
I find some drop in efficiency in the initialStep in the 10mu sample
http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit-pr344-pr345-qCompat_md94a476_c5e29902/plots_building_initialStep/effandfakePtEtaPhi.pdf
(ignore the black histogram there; it's my devel in progress)

mmasciov:
@leonardogiannini had tested this PR few weeks ago. @leonardogiannini, could you post a link?

mkFit/buildtestMPlex.cc
```diff
@@ -639,19 +646,19 @@ void run_OneIteration(const TrackerInfo& trackerInfo, const IterationConfig &itc
       { return StdSeq::qfilter_n_hits_pixseed(t, 3); });
   }
 
-  builder.select_best_comb_cands();
-
   if (do_backward_fit)
   {
-    // a) BackwardFitBH works on MkBuilder::m_tracks
-    builder.BackwardFitBH();
+    builder.BackwardFit();
```
slava77, line 651: this looks like the place that can change the physics results in the initialStep (the rest looks more related to the backward search, which is not used in the initialStep).

@osschar  is my guess correct?
I can make an incremental diff to confirm that some time later.

osschar: Hmmh, could be. They used to be the same ... will cross check the code, we can bisect afterwards if I don't find anything.

slava77: I have a check and a proposal for a  "revert" in #349

PR #348, opened 2021-08-28
add compatibility requirement for propagations to strip hits

slava77:
The following condition is applied on `hit-prop` residuals to make sure that the propagation points consistently within the length bounds of the strip
- `strip hit half-length` + `3*sigma_propagation` > `residual_along_strip` 

More detail:
- the `strip half-length` is related to the uncertainty of its position, in the simplest case of a hit aligned along z, the hit uncertainty along z is `strip half-length` times `sqrt(3)`
- in the barrel the test is simply done in z direction
- in the endcaps the test is done by projecting on the longest component of the hit correlation matrix (taking an approximation that in x-y the uncertainty across the strip (around 100 µm) is negligible.

This is a bit of a kludge until we get the proper geometry details. The CKF tracking has a similar requirement.

Validation using `mkfit=all` and compiled with `AVX2:=1 USE_INTRINSICS:=-DMPT_SIZE=1`
- ttbar http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_mkFit-pr344-pr345-qCompat_m6787d91_c5e29902
- 10mu http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit-pr344-pr345-qCompat_m6787d91_c5e29902/plots_initialStep.html
    - here, for 10mu, the plots show that there is no efficiency or hit loss

In the following, screenshots are for built tracks from the initialStep in the ttbar PU50 sample
orange is the same as the latest physics performance in this PR (6787d91) and has a small improvement compared to black, which was shown during the Aug 27 meeting
image

There is a small reduction in the number of hits due to rejection of fake hits
image
The number of missing outer is a bit better
image

a comparison with CKF is available in http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_CKF_mkFit-pr344-pr345-qCompat_md94a476_c5e29902 (here `qCompat` is as of the earlier version shown in black in the `mkfit=all` comparisons above).

### DIFF
--- a/mkFit/FindingFoos.h
+++ b/mkFit/FindingFoos.h
@@ -9,7 +9,7 @@ class MkBase;
 
 #define COMPUTE_CHI2_ARGS const MPlexLS &,  const MPlexLV &, const MPlexQI &, \
                           const MPlexHS &,  const MPlexHV &, \
-                          MPlexQF &,  const int, const PropagationFlags
+                                MPlexQF &,        MPlexLV &, const int, const PropagationFlags
 
 #define UPDATE_PARAM_ARGS const MPlexLS &,  const MPlexLV &, MPlexQI &, \
                           const MPlexHS &,  const MPlexHV &, \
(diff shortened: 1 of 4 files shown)


slava77:
some time ago we talked about improving `SelectHitIndices` by making corrections to approximately estimate the propagated position with respect to the hit (φ, q). This PR is a post-propagation check of a similar kind in q.
Hopefully this can be eventually moved/reimplemented in the earlier stage in `SelectHitIndices` and save a bit in the propagation cost.

PR #344, opened 2021-08-16
Update phi boundaries in stereo layers, and use floor also in transition region for dphi hit selection windows

mmasciov:
MTV results:
High pT 10μ: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_forPR_10mu/
Low pT 10μ: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_forPR_10mulowpt/
TTbar (PU50): http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_forPR_ttbarPU50/ 

Full documentation: https://docs.google.com/document/d/1YoeWXP4Cvbrrbm2DYV9IeLBnI4XNHPTga4oIgc0IF9w/edit?usp=sharing

PR #349, opened 2021-08-28
back to BackwardFitBH for iters without bw search

slava77:
#345 was not in an obvious way supposed to modify the iterations without the backward search. This PR recovers that behavior.

plots are available for `mkfit=all` option compiled with `AVX2:=1 USE_INTRINSICS:=-DMPT_SIZE=1`
- ttbar http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_mkFit-pr344-pr345-bwNoSearchBH_m7852743_c5e29902
- 10mu http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/10mu_mkFit-pr344-pr345-bwNoSearchBH_m7852743_c5e29902

black (this PR) vs the reference as of #344 are clearly the same based on the ratio plots, e.g. in ttbar initialStep built tracks
image

If the change from #345 for e.g. this `initialStep` case was an improvement, I guess we'd be fine to pick it up. Since this is a regression (a clear increase in fakes), I propose to take this PR and then perhaps investigate if the non-BH solution can be improved (IIUC, it is more flexible).

@osschar 
I've updated both the `run_OneIteration` and `runBtpCe_MultiIter`, but the CMSSW MTV test covers only `run_OneIteration`. 
Please check and if that seems correct, perhaps consider merging.

### DIFF
--- a/mkFit/buildtestMPlex.cc
+++ b/mkFit/buildtestMPlex.cc
@@ -646,19 +646,29 @@ void run_OneIteration(const TrackerInfo& trackerInfo, const IterationConfig &itc
       { return StdSeq::qfilter_n_hits_pixseed(t, 3); });
   }
 
-  if (do_backward_fit)
-  {
-    builder.BackwardFit();
 
-    if (itconf.m_backward_search)
+  if (itconf.m_backward_search)
+  {
+    if (do_backward_fit)
     {
-      builder.BeginBkwSearch();
-      builder.FindTracksCloneEngine(SteeringParams::IT_BkwSearch);
-      builder.EndBkwSearch();
+        builder.BackwardFit();
+
+        builder.BeginBkwSearch();
+        builder.FindTracksCloneEngine(SteeringParams::IT_BkwSearch);
+        builder.EndBkwSearch();
     }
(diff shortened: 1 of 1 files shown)


osschar:
Can you please rather figure out where the difference comes from? We will need this for seed region rebuilding and will need to figure it out anyway.

slava77:
> Can you please rather figure out where the difference comes from? We will need this for seed region rebuilding and will need to figure it out anyway.

OK, I'll give it a try (not before Monday), although I'd guess that you may be more familiar with this part of the code.

if me or someone gets to this before Friday, this PR can wait; otherwise for the next CMSSW update, I'd like to fall back to this revert PR.

slava77:
> OK, I'll give it a try (not before Monday), although I'd guess that you may be more familiar with this part of the code.

it's likely later than Monday. My next item is the pixel bad modules.

osschar:
OK, then it's on  me, I can start looking into this Monday morning. I agree we fall back to this if we can't figure it out in time :)

slava77:
superseded by #351

PR #350, opened 2021-09-01
trivial bug fixes in mkFit.cc

srlantz:
Strictly speaking, the C++ standard requires dimensions of plain arrays to be known at compile time; @dan131riley suggested declaring two arrays as std::vector<double> instead. Other changes are self-explanatory.


mmasciov:
http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/SVA_pr350/

No apparent difference in physics performance: merging.

PR #352, opened 2021-09-03
deadmodules.h extended with bad pixels from 2021 setup

slava77:
this corresponds to `SiPixelQuality_phase1_Run3Beginning` and was generated using trackreco/cmssw#44 .

2 more bad modules was added in `SiPixelQuality_phase1_2021_v1_mc`, which was added/updated somewhat recently in 12X  (the payload was inserted on 2021-07-07).


mmasciov:
As this PR was validated in trackreco/cmssw#44, it should be OK to merge.
@slava77, can you confirm? Thanks.

slava77:
> @slava77, can you confirm?

I confirm

PR #353, opened 2021-09-07
fix error propagation in helixToZ

cerati:
This PR fixes an issue in error propagation in helixToZ.

The effect is to recover the number of strip layers in the high pT 10 mu sample:
http://uaf-10.t2.ucsd.edu/~cerati/plots_Sep05_10mu_testzfix_gcc/plots_building_initialStep/hitsLayers.pdf

This does not have much effect on ttbar.

More plots for both 10mu and ttbar at:
http://uaf-10.t2.ucsd.edu/~cerati/plots_Sep05_10mu_testzfix_gcc/
http://uaf-10.t2.ucsd.edu/~cerati/plots_Sep05_ttbar_testzfix_gcc/

A document with the math I used to derive the jacobian can be found at this link: https://www.overleaf.com/9112147731xmvddqxtwgxq

### DIFF
--- a/mkFit/PropagationMPlex.cc
+++ b/mkFit/PropagationMPlex.cc
@@ -635,6 +635,7 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
       const float cosP = std::cos(phiin), sinP = std::sin(phiin);
       const float cosT = std::cos(theta), sinT = std::sin(theta);
       const float tanT = sinT/cosT;
+      const float icos2T = 1.f/(cosT*cosT);
       const float pxin = cosP*pt;
       const float pyin = sinP*pt;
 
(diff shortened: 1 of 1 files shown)


mkFit/PropagationMPlex.cc
```diff
@@ -666,28 +667,21 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
...
-      errorProp(n,0,5) = -cosP*deltaZ*cosa*sPsCPless1/(cosT*cosT);
-
-      errorProp(n,1,2) = cosa*tanT*(cosP*cosP*sCosPsina - sinP);
-      //      errorProp(n,1,3) = tanT*deltaZ*cosa*( cosP*cosP*sCosPsina + sinP )*pt - k*(sinP*sina + cosP*(1.f-cCosPsina))*pt*pt;
-      errorProp(n,1,3) = tanT*deltaZ*pt*( sinP*cAlessFsAA + cosP*sCPx2 );
-      errorProp(n,1,4) = (k*pt)*( -cosP*sina*sPsCPless1 - 2.f*sinP*sCosPsinah*sCosPsinah );
-      errorProp(n,1,5) = deltaZ*cosa*( cosP*cosP*sCosPsina + sinP )/(cosT*cosT);
+      const float pxcaMpysa = pxin*cosa - pyin*sina;
+      errorProp(n,0,2) = -tanT * ipt * pxcaMpysa;;
+      errorProp(n,0,3) = 2*k*pt*pt*sinah*(sinP*sinah - cosP*cosah) + deltaZ*tanT*pxcaMpysa;
```
slava77, line 672: based on our discussion on skype, this and (1,3) term may need to be refactored to group the large terms (using `deltaZ*tanTheta = k*pt*alpha`)
here `- 2*k*pt*pt*sinah*cosP*cosah = - k*pt*pt*cosP*sina` in the first large term is supposed to [almost] cancel out with the right hand side `deltaZ*tanT*pxin*cosa = k*pt*pt*alpha*cosP*cosa` because `- sina + alpha*cosa` goes to zero; so, it would be best to have this subtraction explicit instead of shuffled with other terms to avoid  loss in numerical precision.

slava77: ... that's for the follow up PR

PR #355, opened 2021-09-09
Correct layer edges

cerati:
This PR corrects the layer edges for the layer spread in r (z) for barrel (endcap).
The implementation is preliminary (currently all in SelectHitIndices and summing linearly the correction to dz or dr), comments are welcome.
The number of layers in the transition region is significantly improved in the high pT 10mu sample:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_10mu_testzfix_edge2_gcc/plots_building_initialStep/hitsLayers.pdf
This comes at a price of some increase in fake rate in ttbar:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfix_edge2_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf
but also an increase in number of layers:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfix_edge2_gcc/plots_building_initialStep/hitsLayers.pdf
Based on the increase in fake rate this may be something we don't want to merge immediately but eventually we should.

It also refactors the terms in the helixAtZ error propagation, for better numerical stability, and removes a double semicolon introduced in my previous PR. 
The effect is small, as expected: http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_10mu_testzfixv2_edge2_gcc/plots_building_initialStep/hitsLayers.pdf

### DIFF
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -294,7 +294,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       float dphi = calcdphi(dphi2, min_dphi);
 
       const float z  = Par[iI].ConstAt(itrack, 2, 0);
-      const float dz = std::abs(nSigmaZ * std::sqrt(Err[iI].ConstAt(itrack, 2, 2)));
+      const float dz = std::abs(nSigmaZ * std::sqrt(Err[iI].ConstAt(itrack, 2, 2))) + std::abs(0.5f*(L.m_layer_info->m_rout-L.m_layer_info->m_rin)/std::tan(Par[iI].ConstAt(itrack, 5, 0)));
       // XXX-NUM-ERR above, Err(2,2) gets negative!
       
       ////// Disable correction
@@ -350,7 +350,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       float dphi = calcdphi(dphi2, min_dphi);
 
       const float  r = std::sqrt(r2);
-      const float dr = nSigmaR*std::sqrt(std::abs(x*x*Err[iI].ConstAt(itrack, 0, 0) + y*y*Err[iI].ConstAt(itrack, 1, 1) + 2*x*y*Err[iI].ConstAt(itrack, 0, 1)) / r2);
+      const float dr = nSigmaR*std::sqrt(std::abs(x*x*Err[iI].ConstAt(itrack, 0, 0) + y*y*Err[iI].ConstAt(itrack, 1, 1) + 2*x*y*Err[iI].ConstAt(itrack, 0, 1)) / r2) + std::abs(0.5f*(L.m_layer_info->m_zmax-L.m_layer_info->m_zmin)*std::tan(Par[iI].ConstAt(itrack, 5, 0)));
 
       ////// Disable correction
       //if (Config::useCMSGeom) // should be Config::finding_requires_propagation_to_hit_pos
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -321,7 +321,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       //  dphi += dist / r;
       //}
 
-      XWsrResult[itrack] = L.is_within_z_sensitive_region(z, dz + edgeCorr);
+      XWsrResult[itrack] = L.is_within_z_sensitive_region(z, std::sqrt(dz*dz + edgeCorr*edgeCorr) );
       assignbins(itrack, z, dz, phi, dphi, min_dq, max_dq, min_dphi, max_dphi);
     }
   }
@@ -370,7 +370,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       //  dphi += std::abs(alpha);
       //}
 
-      XWsrResult[itrack] = L.is_within_r_sensitive_region(r, dr + edgeCorr);
+      XWsrResult[itrack] = L.is_within_r_sensitive_region(r, std::sqrt(dr*dr + edgeCorr*edgeCorr) );
       assignbins(itrack, r, dr, phi, dphi, min_dq, max_dq, min_dphi, max_dphi);
     }
   }
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -294,7 +294,8 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       float dphi = calcdphi(dphi2, min_dphi);
 
       const float z  = Par[iI].ConstAt(itrack, 2, 0);
-      const float dz = std::abs(nSigmaZ * std::sqrt(Err[iI].ConstAt(itrack, 2, 2))) + std::abs(0.5f*(L.m_layer_info->m_rout-L.m_layer_info->m_rin)/std::tan(Par[iI].ConstAt(itrack, 5, 0)));
+      const float dz = std::abs(nSigmaZ * std::sqrt(Err[iI].ConstAt(itrack, 2, 2)));
+      const float edgeCorr = std::abs(0.5f*(L.m_layer_info->m_rout-L.m_layer_info->m_rin)/std::tan(Par[iI].ConstAt(itrack, 5, 0)));
       // XXX-NUM-ERR above, Err(2,2) gets negative!
       
       ////// Disable correction
@@ -320,7 +321,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       //  dphi += dist / r;
       //}
 
-      XWsrResult[itrack] = L.is_within_z_sensitive_region(z, dz);
+      XWsrResult[itrack] = L.is_within_z_sensitive_region(z, dz + edgeCorr);
       assignbins(itrack, z, dz, phi, dphi, min_dq, max_dq, min_dphi, max_dphi);
     }
   }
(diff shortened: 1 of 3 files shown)


mkFit/PropagationMPlex.cc
```diff
@@ -668,14 +668,14 @@ void helixAtZ(const MPlexLV& inPar,  const MPlexQI& inChg, const MPlexQF &msZ,
 		<< " pxin=" << pxin << " pyin=" << pyin);
 
       const float pxcaMpysa = pxin*cosa - pyin*sina;
-      errorProp(n,0,2) = -tanT * ipt * pxcaMpysa;;
-      errorProp(n,0,3) = 2*k*pt*pt*sinah*(sinP*sinah - cosP*cosah) + deltaZ*tanT*pxcaMpysa;
+      errorProp(n,0,2) = -tanT * ipt * pxcaMpysa;
+      errorProp(n,0,3) = k*pt*pt*( cosP*(alpha*cosa - sina) + sinP*( 1.f - cosa - alpha*sina) );
```
slava77: please regroup the second factor to use alpha/2
`1.f - cosa - alpha*sina = 2.f*sinah*sinah - alpha*2.f*sinah*cosah = 2.f*sinah*(sinah - alpha*cosah)`
this will avoid `1 -1` kind of subtraction

mkFit/MkFinder.cc
```diff
@@ -350,7 +350,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
       float dphi = calcdphi(dphi2, min_dphi);
 
       const float  r = std::sqrt(r2);
-      const float dr = nSigmaR*std::sqrt(std::abs(x*x*Err[iI].ConstAt(itrack, 0, 0) + y*y*Err[iI].ConstAt(itrack, 1, 1) + 2*x*y*Err[iI].ConstAt(itrack, 0, 1)) / r2);
+      const float dr = nSigmaR*std::sqrt(std::abs(x*x*Err[iI].ConstAt(itrack, 0, 0) + y*y*Err[iI].ConstAt(itrack, 1, 1) + 2*x*y*Err[iI].ConstAt(itrack, 0, 1)) / r2) + std::abs(0.5f*(L.m_layer_info->m_zmax-L.m_layer_info->m_zmin)*std::tan(Par[iI].ConstAt(itrack, 5, 0)));
```
slava77: linebreaks would help to view this more clearly on the web

slava77:
@cerati 
are both barrel and endcap extensions to `dq` useful?
Most of our discussion was about the boundaries in the endcaps.

It's a bit surprising that this rather general update is needed in `SelectHitIndices`, which is supposed to be already covered by the window tuning which was done per layer and parameterized in eta.

Can this update be restricted only to the sensitive region (`XWsrResult`) computation and not include it in the (φ,q) bin computation? Perhaps this will reduce the fakes.

mkFit/MkFinder.cc
slava77: more specific on what I meant in https://github.com/trackreco/mkFit/pull/355#issuecomment-916538932

```suggestion
      const float dr = nSigmaR*std::sqrt(std::abs(x*x*Err[iI].ConstAt(itrack, 0, 0) + y*y*Err[iI].ConstAt(itrack, 1, 1) + 2*x*y*Err[iI].ConstAt(itrack, 0, 1)) / r2);
      const float dr_for_wsr = dr + std::abs(0.5f*(L.m_layer_info->m_zmax-L.m_layer_info->m_zmin)*std::tan(Par[iI].ConstAt(itrack, 5, 0)));
```
and then  below
```diff
-       XWsrResult[itrack] = L.is_within_r_sensitive_region(r, dr);
+      XWsrResult[itrack] = L.is_within_r_sensitive_region(r, dr_for_wsr);
```

cerati: good point... I missed the fact that dr is not used just for WSR...

cerati:
new errorProp factorization (v3):
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_10mu_testzfixv3_edge2_gcc/plots_building_initialStep/hitsLayers.pdf

using edge correction only for WSR:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev3_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf

In both cases there are negligible differences wrt before.

slava77:
@cerati 
can you check if a restriction to add a `-3` hit only if there are no hits on the layer is going to work better?
it would be similar to https://github.com/trackreco/mkFit/blob/devel/mkFit/MkFinder.cc#L1143-L1148
```diff
-    if (XWsrResult[itrack].m_wsr == WSR_Edge)
+    if (XWsrResult[itrack].m_wsr == WSR_Edge && nHitsAdded[itrack] == 0)
    {
      fake_hit_idx = -3;
```

mkFit/MkFinder.cc
```diff
@@ -295,6 +295,7 @@ void MkFinder::SelectHitIndices(const LayerOfHits &layer_of_hits,
 
       const float z  = Par[iI].ConstAt(itrack, 2, 0);
       const float dz = std::abs(nSigmaZ * std::sqrt(Err[iI].ConstAt(itrack, 2, 2)));
+      const float edgeCorr = std::abs(0.5f*(L.m_layer_info->m_rout-L.m_layer_info->m_rin)/std::tan(Par[iI].ConstAt(itrack, 5, 0)));
```
slava77, line 298: it's worth to check if using `r/z` is going to work better here instead of tanTheta. This would avoid promoting significantly "twisted" tracks from being promoted to the edge category (e.g. an update with a hit far away in z will align the track along z and make this correction very large).

cerati: are you referring to track r/z or layer r/z? I tried the latter in the first version (commit 8b69bbc) with larger fake rates (http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfix_edge_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf)
So I am not very optimistic. Also with the commit below the fake rate increase is further reduced

slava77: I meant the track r/z

slava77:
Perhaps this is already obvious, but still.

as seen in one of the recently shared fr plots 
image
the relative increase blows up just below 0.8 GeV. The first bin 0.65-0.8 and the bins below include tracks which turn around and loop back (0.71 GeV has max radius of 125 cm [inner edge of ECAL]; 0.63 GeV has max radius of 110 cm [outer extent of the tracker]).

- On one hand we should probably not collect any hits for tracks with pt > 0.7 GeV that are propagated back. Where is the stopping logic in the code?
- on the other hand, for the lower pt tracks, is error (and what else) propagation code debugged already for the part of the trajectory going inward in xy?

cerati:
Hi @slava77, 

The last commit adds the edge correction and the dz uncertainty in quadrature, and this significantly reduces the increase in fake rate since the uncertainty has a sizable contribution for low pt tracks. The validation plots can  be found at this link:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_10mu_testzfixv3_edgev4_gcc/
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev4_gcc/
Those that are most relevant are:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev4_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev4_gcc/plots_building_initialStep/hitsLayers.pdf
You can compare with the previous version (edgev3):
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev3_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev3_gcc/plots_building_initialStep/hitsLayers.pdf

I also tried to add a -3 hit only if there are no hits on the layer, as you suggested. This is v5:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev5_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev5_gcc/plots_building_initialStep/hitsLayers.pdf
As you can see there is a significant reduction in number of layers for ttbar (although not for 10mu), but not a big reduction in fake rate. So at the moment I think v4 is the best overall.

slava77:
> The last commit adds the edge correction and the dz uncertainty in quadrature, and this significantly reduces the increase in fake rate since the uncertainty has a sizable contribution for low pt tracks. The validation plots can be found at this link:

it's pretty hard to follow the differences between v3, v4, and v5.
Please make a set of plots that compares the three.
Thank you.

slava77:
> I also tried to add a -3 hit only if there are no hits on the layer, as you suggested. This is v5:
> http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev5_gcc/plots_building_initialStep/effandfakePtEtaPhi.pdf
> http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_testzfixv3_edgev5_gcc/plots_building_initialStep/hitsLayers.pdf
> As you can see there is a significant reduction in number of layers for ttbar (although not for 10mu), but not a big reduction in fake rate. So at the moment I think v4 is the best overall.

isn't this more suggestive that we are adding fake hits?

cerati:
> isn't this more suggestive that we are adding fake hits?
not necessarily, since the fake rate does not change much

cerati:
> Please make a set of plots that compares the three.

http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_edgev3v4v5_gcc/plots_initialStep.html

slava77:
> > Please make a set of plots that compares the three.
> 
> http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep08_ttbar_edgev3v4v5_gcc/plots_initialStep.html

v4->v5: 
- the reduction in the fake rate is quite significant in the transition region and TEC edge (eta 1.4 and 2.5); here v5 is better.
- the reduction in the number of strip layers in the transition region is perhaps still OK, but the drop at the edge of the TEC is fractionally much larger; here v5 is worse.

So, I wouldn't call v4 a clear winner, but it is perhaps OK to go on with v4.

slava77:
@cerati 
were you going to do more checks or is this ready to merge?

cerati:
@slava77 I am not doing more checks. I guess we should tackle the low pt fakes independently from this PR, so it's probably ok to merge it despite it somehow increase the fake rate

PR #358, opened 2021-09-16
Cluster size cuts

cerati:
Add option for cluster row size cut (iteration and layer-dependent), and turn it on for InitialStep.

Tested on initialStep within MTV:
http://uaf-8.t2.ucsd.edu/~cerati/plots_Sep16_ttbar_clshcut_PR/plots_initialStep.html

A PR to CMSSW with JSON updates will also be needed.

PR #360, opened 2021-09-21
Stop candidates before they reach apogee.

osschar:
Stops tracks when their dr/ds angle gets shallower than 0.2 rad (11.5 deg).
Radius below which the stopping is not applied is 25cm. 
There is no pT cut, it does not seem to be needed, but could save us some atan2s if we apply it and the r_min cut first.

I don't know if tuning of dr/ds cut makes sense.

I suspect this will partially overlap with large-cluster cut in #358, esp. in barrel

MTV vs CKF and #355 (devel): http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/stop-loopers
MTV with #355 as ref:  http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/stop-loopers.vs.355
```
sample=ttbarpu50
maxEvents=1000
mkfit="InitialStepPreSplitting,InitialStep,HighPtTripletStep,DetachedQuadStep,DetachedTripletStep,PixelLessStep,TobTecStep"
```

### DIFF
--- a/mkFit/MkBuilder.cc
+++ b/mkFit/MkBuilder.cc
@@ -1467,7 +1467,8 @@ void MkBuilder::find_tracks_load_seeds(const TrackVec &in_seeds)
 
 int MkBuilder::find_tracks_unroll_candidates(std::vector<std::pair<int,int>> & seed_cand_vec,
                                              int start_seed, int end_seed,
-                                             int prev_layer, bool pickup_only)
+                                             int layer, int prev_layer, bool pickup_only,
+                                             SteeringParams::IterationType_e iteration_dir)
 {
   int silly_count = 0;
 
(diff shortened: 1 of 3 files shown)


slava77:
@osschar 
please also run on 10muvlowpt sample (this is available directly in the input options since https://github.com/trackreco/cmssw/pull/47 was merged on Sep 17)

osschar:
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/mu-stop-loopers/
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/mu-stop-loopers.vs.355/

mkFit/MkBuilder.cc
```diff
@@ -1488,6 +1488,15 @@ int MkBuilder::find_tracks_unroll_candidates(std::vector<std::pair<int,int>> & s
       {
         if (ccand[ic].getLastHitIdx() != -2)
         {
+         // Check if the candidate is close to it's max_r, pi/2 - 0.2 rad (11.5 deg)
+         const float dphi = std::abs(ccand[ic].posPhi() - ccand[ic].momPhi());
```
slava77: will there be a problem at +/-π?

osschar: see one line down

slava77: ah, OK

slava77:
are there any concerns to merge this?
(just as a heads up check before doing it probably after the meeting within an hour)

cerati:
> are there any concerns to merge this?
> (just as a heads up check before doing it probably after the meeting within an hour)

is this taking precedence over #357 and #358?

osschar:
i want to add pt cut (1.2) and forward_search check ... and only calculate dphi if all else is true.
e might weant to disable it for some iterations, i don't know.
there is the question if we should fiddle with delta_phi, now hardcoded to 11.5 deg.

slava77:
> is this taking precedence over #357 and #358?

I think that this is independent

osschar:
I'll just rerun ttbar validation to make sure i didn't screw up something.

osschar:
Comparing stop-loopers before / after adding of pT cut -- practically all eff/fakes for built-tracks flat.
http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/stop-loopers-1.vs.2/

PR #361, opened 2021-09-25
reapply strip cluster charge cut (loose) using track impact angle

slava77:
- the strip cluster charge is expected to scale proportionally to the track path in the module
- the raw `pcm = charge/L_normal` is applied upstream in CMSSW gives the loosest possible cut cluster charge
- this PR computes a corrected `qCorr = charge/L_path = charge/(L_normal*p/p_zLocal) = pcm*p_zLocal/p`, where the zLocal direction is orthogonal to the module plane
    - hits with `pcm` at or above the overflow value 3652 ("only" 2.2 times the cut value) are not corrected
    - in the barrel it is assumed/approximated to be in the (x,y) plane and the `p_zLocal` projection is made using the hit error matrix component in the (x,y) plane. The math is expressed in vector terms without explicit eigenvector computation. Here a diagonalized version is expected to be `diag(sigma_strip^2, 0)`, or `proj = msErr/trace(msErr)` after diagonalization to be `diag(1, 0)` within numerical precision. About 1% of the hits is not corrected if they have `det(proj)> 0.1`, which would mean that the module normal is partly along the global z direction due to misalignment. For the remaining majority of the hits `qSF = p_zLocal/p = sqrt[(px,py)*(1-proj)*(px,py)]/p = sqrt[(cosP,sinP)*(1-proj)*(cosP,sinP)]`
    - in the endcap the zLocal is simply assumed to be the global z direction

The reason to apply this cut was motivated by the tracker DQM plots https://tinyurl.com/yg3qvray  
image
here in mkFit we start from the cut at 1620 on `pcm` and the plot is made for relatively straight tracks (`p>2 && pt>0.8`), so after the correction to the origin the hits can extend to 1620/[cosh(η_max for TIB1 = 1.7) = 2.8] ≈ 600`, in pretty good agreement with the plot.

Validation
- mtv-7iter http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_CKF_mkFit-7-pr357_PCMfromTrack_mffe5c6e_c9266151/
- 4-iter http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_CKF_mkFit-4-pr357_PCMfromTrack_mffe5c6e_c9266151/

In the 7iter initial step built tracks:
the fakerate at low pt is down by about 5% relative with a small increase in efficiency
image
image
only strip hits count changes (pixels are unchanged, as expected) and mostly in the η range from 1 to 1.7 where the angle wrt module normal is the largest
image

all tracks:
image

The CPU cost seems to be small: https://github.com/trackreco/mkFit/pull/361#issuecomment-927323163 suggests that the increase if any is below 1% of the building time

After this PR:
- on a longer term, perhaps compare to a correction assuming straight line from origin, using the hit position; this should save quite a bit of CPU cost by avoiding 3 sin/cos computations
- consider a configurable cut per iteration (some iters have a tight cut at 1945 (instead of a loose value at 1620)); this would allow a tighter cut in some places, compared to this hardcoded baseline

Things to check before merging
- check if the tracker DQM confirms that the updated charge distributions now agree with CKF (@mmasciov do you have a setup for this?)

### DIFF
--- a/Hit.h
+++ b/Hit.h
@@ -278,6 +278,7 @@ public:
   unsigned int spanRows()     const { return pdata_.span_rows + 1; }
   unsigned int spanCols()     const { return pdata_.span_cols + 1; }
 
+  static unsigned int minChargePerCM() { return 1620; }
   static unsigned int maxChargePerCM() { return 1620 + (0xfe << 3); }
   static unsigned int maxSpan()        { return 8; }
 
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -842,6 +842,42 @@ bool isStripQCompatible(int itrack, bool isBarrel, const MPlexLS &pErr,  const M
   }
 }
 
+//=======================================================
+// passStripChargePCMfromTrack : apply the slope correction to charge per cm and cut using hit err matrix
+//         the raw pcm = charge/L_normal
+//         the corrected qCorr = charge/L_path = charge/(L_normal*p/p_zLocal) = pcm*p_zLocal/p
+//=======================================================
+bool passStripChargePCMfromTrack(int itrack, bool isBarrel, unsigned int pcm, unsigned int pcmMin,
+                                 const MPlexLV &pPar, const MPlexHS &msErr)
+{
+  //skip the overflow case
+  if (pcm >= Hit::maxChargePerCM())
+    return true;
+
+  float qSF;
+  if (isBarrel) {//project in x,y, assuming zero-error direction is in this plane
+    const float hitT2 = msErr.ConstAt(itrack,0,0) + msErr.ConstAt(itrack,1,1);
+    const float hitT2inv = 1.f/hitT2;
+    const float proj[3] = {msErr.ConstAt(itrack,0,0)*hitT2inv, msErr.ConstAt(itrack,0,1)*hitT2inv,
+                           msErr.ConstAt(itrack,1,1)*hitT2inv};
+    const bool detXY_OK = std::abs(proj[0]*proj[2] - proj[1]*proj[1]) < 0.1f;//check that zero-direction is close
+    const float cosP = cos(pPar.ConstAt(itrack,4,0));
+    const float sinP = sin(pPar.ConstAt(itrack,4,0));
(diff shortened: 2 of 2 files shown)


slava77:
timing doesn't seem to be be visibly affected, based on 100 event test with 1 thread:
http://uaf-10.t2.ucsd.edu/~slava77/figs/mic/mtv/ttbar50_mkFit-7-pr357_PCMfromTrack_dqm-1_mffe5c6e_c9266151/plots_timing/iterationsReal.pdf
there is about 1-2% increase in building time, but the seeding time is also up by about the same. So, this looks more like a general float of the process

slava77:
I've updated the "todo" list in the PR description to 

#

The CPU cost seems to be small: https://github.com/trackreco/mkFit/pull/361#issuecomment-927323163 suggests that the increase if any is below 1% of the building time

After this PR:
- on a longer term, perhaps compare to a correction assuming straight line from origin, using the hit position; this should save quite a bit of CPU cost by avoiding 3 sin/cos computations
- consider a configurable cut per iteration (some iters have a tight cut at 1945 (instead of a loose value at 1620)); this would allow a tighter cut in some places, compared to this hardcoded baseline

Things to check before merging
- check if the tracker DQM confirms that the updated charge distributions now agree with CKF (@mmasciov do you have a setup for this?)

#

mmasciov:
@slava77 Yes, I can submit this for TTbar right away, after merging locally your PR

mmasciov:
@slava77 : difference wrt. CKF is indeed largely reduced after this PR.
- mkFit 4-iter: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/TrackerDPGValidation/plotsDQM_TrackerDPG_mkFit-4iter_PR361/Summary_ClusterChargePerCMfromOrigin__OnTrack__TIB__layer__1.png
- mkFit 7-iter: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/TrackerDPGValidation/plotsDQM_TrackerDPG_mkFit-7iter_PR361/Summary_ClusterChargePerCMfromOrigin__OnTrack__TIB__layer__1.png

PR #363, opened 2021-10-07
Add BeamSpot.

osschar:
Please check default values for sigma and beam-width.

Read/Write of BeamSpot for standalone bin files still needs to be implemented.

### DIFF
--- a/mkFit/MkStdSeqs.cc
+++ b/mkFit/MkStdSeqs.cc
@@ -15,7 +15,7 @@ namespace StdSeq {
 // Hit processing
 //=========================================================================
 
-void LoadHits(Event &ev, EventOfHits &eoh)
+void LoadHitsAndBeamSpot(Event &ev, EventOfHits &eoh)
 {
     eoh.Reset();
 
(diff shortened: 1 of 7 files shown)


mkFit/MkStdSeqs.cc
```diff
@@ -15,7 +15,7 @@ namespace StdSeq {
 // Hit processing
 //=========================================================================
 
-void LoadHits(Event &ev, EventOfHits &eoh)
+void LoadHitsAndBeamSpot(Event &ev, EventOfHits &eoh)
```
slava77, line 18: CMSSW is not using this part; `Cmssw_LoadHits_Begin` is used.
perhaps a separate method that can pass the beamspot is easier; it will be used as `LoadDeads`

osschar: Yes, cmssw will just call eoh.SetBeamSpot().
I thought it's cleaner to have a single method for standalone and rename it express intent.

slava77: ok

slava77:
I can imagine that the small test I did in trackreco/cmssw#54 is enough to confirm that this can be merged.

Any objections?

osschar:
Yes, looks good ... mayeb Mario can push his addition to this branch as well?

slava77:
> Yes, looks good ... mayeb Mario can push his addition to this branch as well?

I think that we can easily decompose to separate PRs.

PR #365, opened 2021-10-08
Relax detached triplet filter

mmasciov:
### PR description

Updated (WP of the) detachedTripletStep quality filter is aimed at improving "Built track" efficiency as much as possible, with ~ no increase in fake rate. No change is expected in other iterations.

It requires update of mkFit JSON files: trackreco/cmssw#55

### PR validation (PR #365 + trackreco/cmssw#55):

Expectation described above is confirmed by MTV results collected below.
--> PRs are ready to be merged (in absence of objections).

(1) NO change in mkFit-4iter: 
http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/SVA_displacedIters/detachedTripletFilter/MTV_mkFit-4iter_PR365/

(2) Slight increase in efficiency up to "All tracks" for mkFit-7iter, due to increase in efficiency for detachedTripletStep: http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/SVA_displacedIters/detachedTripletFilter/MTV_mkFit-7iter_PR365_all/

(3) When mkFit is enabled only in detachedTripletStep, improvement in efficiency up to "All tracks": 
http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/SVA_displacedIters/detachedTripletFilter/MTV_mkFit-detachedTripletStep_PR365_all/

(4) When mkFit is enabled in 4-iter + detachedTripletStep, improvement in efficiency up to "All tracks":
http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/SVA_displacedIters/detachedTripletFilter/MTV_mkFit-4iter-plusDetachedTripletStep_PR365_all/

PR #364, opened 2021-10-08
write/read beam spot in a standalone setup

slava77:
The file version is updated to v6.
The read/write code should be backward compatible with the older versions 4 (before the masks) and 5.

I made a test file `11834.0_TTbar_14TeV+2021/AVE_50_BX01_25ns/memoryFile.fv6.default.211007-652f30c.bin`
but I did not try to run the standalone validation on this yet.

PR #366, opened 2021-10-09
update bin files  fv6.default.211008-c6b7c67.bin to include beamSpot information

slava77:
This is a catch-up PR for #364 

from from fv5.default.210729-d1dc487.bin (full stats)
to fv6.default.211008-c6b7c67.bin (limited stats: 1K ttbar; 10K evt particle guns).

I did not run standalone validation on these: 
- the new feature ( beamSpot) will be used only in #365 
- I expect that the simple reduction in stats should be manageable at this stage (except for some older full node long throughput tests, we never used so many events)


mmasciov:
I have run a local test on TTbar PU50, and I see no issue.

PR #369, opened 2021-10-12
Update pixelLessStep duplicate removal and add filter to reduce fakes

mmasciov:
### PR validation:

https://www.dropbox.com/s/ww1chjzsrnseklv/mkFit_additionalFilter_pixelLessStep.pdf?dl=0

### DIFF
--- a/mkFit/MkStdSeqs.h
+++ b/mkFit/MkStdSeqs.h
@@ -67,6 +67,17 @@ namespace StdSeq
 		 (layers<=6 && lastInsidePix && llyr!=lplyr && pt<=pt_minH && std::abs(d0BS)>d0_max) );
     }
 
+    template<class TRACK>
+    bool qfilter_n_layers_pixelLess(const TRACK &t, const BeamSpot &bspot)
+    {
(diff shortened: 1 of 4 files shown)


mkFit/MkStdSeqs.h
```diff
@@ -67,6 +67,17 @@ namespace StdSeq
...
 
+    template<class TRACK>
+    bool qfilter_n_layers_pixelLess(const TRACK &t, const BeamSpot &bspot)
+    {
+      int layers   = t.nUniqueLayers();
+      int nhits    = t.nFoundHits();
+      float d0BS   = t.d0BeamSpot(bspot.x,bspot.y);
+      float d0_max = 0.1; // 1 mm
+
+      return !( (nhits<=6 || layers<=6) && std::abs(d0BS)>d0_max );
```
slava77, line 78: can we check also if counting rphi (strip) layers specifically would help.
The track DNN does not distinguish and we could gain here more explicitly.

in the TrackerInfo notations, it is to count layers `(! is_pix_lyr) && (!is_stereo_lyr))`

PR #370, opened 2021-10-20
Adding counting of layers with match to TrackCand

leonardogiannini:
Adding a function to track cand that counts once the layers if they are a mono+stereo pair and they both have a valid hit 

Example of matching and counting reduction in the figure going from 5 to 3 layers. the hits are counte inwards and the stereo layer is the outer one

image

this counter should be tested in filters that we already are using

### DIFF
--- a/mkFit/HitStructures.h
+++ b/mkFit/HitStructures.h
@@ -512,8 +512,14 @@ public:
   int        getLastFoundHitLyr() const;
   int        nUniqueLayers()      const;
 
-  int        nUniqueLayersMatch(const TrackerInfo &trk_inf) const;
-  int        nLayersByType(const TrackerInfo &trk_inf) const;
+  int        nLayersByTypeEncoded(const TrackerInfo &trk_inf) const;
+  int        nHitsByTypeEncoded(const TrackerInfo &trk_inf) const;
+
+  int        nPixelDecoded(const int &encoded)        const {return encoded%100;} 
+  int        nStereoDecoded(const int &encoded)       const {return (encoded/100)%100;} 
+  int        nMonoDecoded(const int &encoded)         const {return (encoded/10000)%100;} 
+  int        nMatchedDecoded(const int &encoded)      const {return encoded/1000000;}
+  int        nTotMatchDecoded(const int &encoded)     const {return encoded%100+(encoded/100)%100+(encoded/10000)%100-encoded/1000000;} 
 
   void  addHitIdx(int hitIdx, int hitLyr, float chi2);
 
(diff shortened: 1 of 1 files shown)


leonardogiannini:
I added a counter of Layers in groups of pixel / stereo / mono (non stereo)
I tried to avoid calling the function three times for the same loop, so the result is in the format __ - __ - __ (e.g. 05-02-01)
I can change the format if necessary

mmasciov:
> I added a counter of Layers in groups of pixel / stereo / mono (non stereo) I tried to avoid calling the function three times for the same loop, so the result is in the format __ - __ - __ (e.g. 05-02-01) I can change the format if necessary

@leonardogiannini, if I understand correctly the sum of pix+mono+stereo equals nUniqueLayers, which is currently computed by a different function.
Maybe it would be better to just have a single function that returns all layer counters?
That is, a function that:
- returns pix+mono+stereo (=nUniqueLayers);
- takes nPixelLayers, nStripMonoLayers, nStripStereoLayers and nStripMatchedLayers as arguments, and assigns a value to each of those within the same loop.

I guess this would limit the number of loops in case more counters are needed?

On the other hand, I believe it would be good to also have analogous counters for the number of hits: nPixelHits, nStripMonoHits, nStripStereoHits, and maybe nStripGluedHits (corresponding to nStripMatchedLayers)?

leonardogiannini:
I tried to write a function that updates the values given as arguments, but I had a strange compilation error
the object has cv-qualifiers that are not compatible with the member function "mkfit::TrackCand::nLayersByType"

for the hits I can write the function. the only doubt is what to do when we have 4 hits (2 doubles) in glued layers. It's very rare, but I saw this double overlap.

mkFit/HitStructures.h
```diff
@@ -782,6 +791,72 @@ inline int TrackCand::nUniqueLayers() const
...
+  int nh = nTotalHits();
+  int ch = lastHitIdx_;
+  int pix=0, stereo=0, mono=0, matched=0;
+  while (--nh >= 0)
+  {
+    HoTNode& hot_node = m_comb_candidate->m_hots[ch];
+    int thisL = hot_node.m_hot.layer;
+    if (thisL >=0 && (hot_node.m_hot.index >= 0 || hot_node.m_hot.index == -9) && thisL != prevL)
+    {
+      std::cout << "thisL " << thisL <<" p "<< trk_inf.is_pix_lyr(thisL) <<" s "<< trk_inf.is_stereo_lyr(thisL) << std::endl;
```
mmasciov: Can this be commented out?
After this, I think this is ready to be merged.

leonardogiannini: yes, I forgot it

Issue #373, opened 2021-10-24
filter out candidates with nans and negative diagonal covariance elements


slava77:
this is a follow-up to cms-sw/cmssw#35798 

unlike in the standalone setup, the `run_OneIteration` used in cmssw is missing `check_nan_n_silly_*`.
This issue is to propose to add these checks and to track the progress.

PR #374, opened 2021-10-25
Add filter to remove nan and silly candidates

mmasciov:
### PR description
This PR introduces a filter on 'silly' candidates, to avoid error/warning messages in CMSSW, addressing #373.

### PR validation
Physics performance is unchanged.
Timing performance is also ~ unchanged (in the plots linked below there is an offset in the measurement; however, after subtracting such offset, timing is indeed unchanged).
MTV results based on TTbar (<PU>=50): http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_testNanNSillyRemoval/
Running on 100 events, the size of the log file decreases from 3.6 MB to 1.3 MB 
(note that the warning suppression introduced by Slava in cms-sw/cmssw#35802 is not included).
The instances of 'candidate ignored' decrease from 4778 to 53.

### DIFF
--- a/Track.cc
+++ b/Track.cc
@@ -183,11 +183,8 @@ bool TrackBase::hasNanNSillyValues() const
     {
       if ((i == j && state_.errors.At(i,j) < 0) || ! std::isfinite(state_.errors.At(i,j)))
       {
-        if ( ! is_silly )
-        {
-          is_silly = true;
-	  break;
-        }
+	is_silly = true;
+	return is_silly;
       }
     }
   }
--- a/Track.cc
+++ b/Track.cc
@@ -160,7 +160,7 @@ bool TrackBase::hasSillyValues(bool dump, bool fix, const char* pref)
     {
       if ((i == j && state_.errors.At(i,j) < 0) || ! std::isfinite(state_.errors.At(i,j)))
       {
-        if ( ! is_silly)
+        if ( ! is_silly )
         {
           is_silly = true;
           if (dump) printf("%s (label=%d, pT=%f):", pref, label(), pT());
--- a/Track.cc
+++ b/Track.cc
@@ -174,6 +174,26 @@ bool TrackBase::hasSillyValues(bool dump, bool fix, const char* pref)
   return is_silly;
 }
 
+bool TrackBase::hasNanNSillyValues() const
+{
+  bool is_silly = false;
+  for (int i = 0; i < LL; ++i)
+  {
+    for (int j = 0; j <= i; ++j)
+    {
+      if ((i == j && state_.errors.At(i,j) < 0) || ! std::isfinite(state_.errors.At(i,j)))
+      {
+        if ( ! is_silly )
+        {
+          is_silly = true;
+	  break;
+        }
+      }
+    }
+  }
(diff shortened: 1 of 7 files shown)


osschar:
Filter-style removal makes sense between forward-search and backward fit. It's somewhat expensive as it has to move comb-cands up in the vector once one of them gets removed. We do run all filters at the same time, don't we?

I thought we need to to do this at the very end ... so why not just have the check in export function ... and simply not export the CombCand for which this fails.

How about seeds? Do we want to check there, too? There was a "fix" option (that didn't make much sense to me and I remember asking what should actually be done with those seeds / or what beter values to put there would be / or where this would need to be fixed upstream (in the seed producer)).

mmasciov:
> Filter-style removal makes sense between forward-search and backward fit. It's somewhat expensive as it has to move comb-cands up in the vector once one of them gets removed. We do run all filters at the same time, don't we?
> 
> I thought we need to to do this at the very end ... so why not just have the check in export function ... and simply not export the CombCand for which this fails.
> 
> How about seeds? Do we want to check there, too? There was a "fix" option (that didn't make much sense to me and I remember asking what should actually be done with those seeds / or what beter values to put there would be / or where this would need to be fixed upstream (in the seed producer)).

This is done at the very end, after backward fit:
https://github.com/trackreco/mkFit/pull/374/files#diff-9bd117b4474e1261389e0f33d9f898dc985c780ab0e7c9c12339ce68867a0e66R715
There are also new filters applied at the end.
It could probably be done also during the export, I guess.

As for seeds, currently we assign an error of 1e-5.
We could try to remove the bad seeds instead (there is an option, disabled by default, which can remove silly seeds in the seed post-processing, instead if 'fixing' them.

osschar:
OK, then let's leave the final filter optimization for next time (maybe passing a vector of filter functions to export) :)

I don't know what to do with bad seeds ... and how many of them there even are in each iteration.

mmasciov:
> OK, then let's leave the final filter optimization for next time (maybe passing a vector of filter functions to export) :)
> 
> I don't know what to do with bad seeds ... and how many of them there even are in each iteration.

I have enabled by default the removal of silly seeds with commit 9ef01f6, if any. No difference, as this is very unlikely to happen:
http://uaf-10.t2.ucsd.edu/~mmasciov/MkFit_devs/MTV_testNanNSillyRemoval_sillySeedRemoval/

Track.cc
```diff
@@ -174,6 +174,26 @@ bool TrackBase::hasSillyValues(bool dump, bool fix, const char* pref)
...
+{
+  bool is_silly = false;
+  for (int i = 0; i < LL; ++i)
+  {
+    for (int j = 0; j <= i; ++j)
+    {
+      if ((i == j && state_.errors.At(i,j) < 0) || ! std::isfinite(state_.errors.At(i,j)))
+      {
+        if ( ! is_silly )
+        {
```
slava77: why is the `if ( ! is_silly )` needed here?
should it just return here instead of continuing to loop over `i`

mmasciov: Yes, it's right. Pushing an update.

PR #376, opened 2021-10-26
Do not export missing hits, avoid extra material effect application in backward-fit

osschar:
- For clone-engine keep candidate state at last valid hit.
- Add option to export candidates without missing hit information, use it for CMSSW runOneIteration().
- In backward fit avoid re-application of material effects when a track does not have a valid hit on current layer.

MTV with PR-372 as ref: http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/no_missing_hits_und_no_mateff_bkfit.vs.PR-372
MTV with CKF as ref: http://xrd-cache-1.t2.ucsd.edu/matevz/PKF/no_missing_hits_und_no_mateff_bkfit.vs.CKF

### DIFF
--- a/mkFit/HitStructures.cc
+++ b/mkFit/HitStructures.cc
@@ -445,27 +445,33 @@ EventOfHits::EventOfHits(const TrackerInfo &trk_inf) :
 // TrackCand
 //==============================================================================
 
-Track TrackCand::exportTrack() const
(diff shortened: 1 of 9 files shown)


mmasciov:
After fixing a minor conflict, merging this PR.

PR #383, opened 2021-11-20
Builder release memory per event

osschar:
Memory allocated for TrackCand vectors is released on every event / iteration -- before it was growing to accommodate the largest event seen so far.

Further, overlap hit-storage vector that was member of CombinedCandidate has been moved as member variable in TrackCand -- thus it does not need to be allocated as a vector per seed anymore.

igProf MEM_LIVE 50 ttbar pu 50 events:
http://xrd-cache-1.t2.ucsd.edu/matevz/igProf/igprof-navigator.py/IgProf.50

DQM file for 7 iterations, ttbar 50 isin phi3:/data2/matevz/CMSSW_11_2_0/src/mem_test/ ... I don't have relevant reference files on hand.


osschar:
@makortel can you please check?

makortel:
As far as I can tell (including the IgProf report), looks good!

slava77:
> As far as I can tell (including the IgProf report), looks good!

I merged based on this

PR #385, opened 2021-11-23
Fix lower dphi window bound (affecting pixelLess and detachedTriplet)

mmasciov:
### PR description:
As per title, this fixes a bug in the setting of hit selection dphi windows.
Additionally, it adds an upper bound on 1/pT for window parametrization, for both dq and dphi, to avoid that window size explodes if pT<0.1 GeV.
These issues seem to only have affected detachedTripletStep and pixelLessStep, in terms of timing.
Effects on physics performance are ~ invisible (see below).

### PR validation:
TTbar <PU>=50: https://mmasciovecchio.web.cern.ch/MkFit_hitSelectionWindows_Nov21/MTV_fixHitSelWindows_Nov21_TTbarPU50_122X/

10muPt0p2to10: https://mmasciovecchio.web.cern.ch/MkFit_hitSelectionWindows_Nov21/MTV_fixHitSelWindows_Nov21_10muPt0p2to10_122X/

"Luckily", changes in physics performance are ~ invisible.
On the other hand, timing for pixelLessStep and detachedTripletStep appears to be "fixed":
https://mmasciovecchio.web.cern.ch/MkFit_hitSelectionWindows_Nov21/MTV_fixHitSelWindows_Nov21_TTbarPU50_122X/plots_timing/iterationsReal.pdf

### DIFF
--- a/mkFit/MkFinder.cc
+++ b/mkFit/MkFinder.cc
@@ -167,8 +167,11 @@ void MkFinder::getHitSelDynamicWindows(const float invpt, const float theta, flo
 {
   const IterationLayerConfig &ILC = *m_iteration_layer_config;
 
+  float max_invpt=invpt;
+  if(invpt>10.0) max_invpt=10.0; // => pT>0.1 GeV
+
   // dq hit selection window
-  float this_dq = (ILC.c_dq_0)*invpt+(ILC.c_dq_1)*theta+(ILC.c_dq_2);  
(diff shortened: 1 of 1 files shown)


leonardogiannini:
examples of timing for a few events from /eos/cms/store/user/cmsbuild/profiling/data/CMSSW_12_2_0_pre2/slc7_amd64_gcc900/11834.21/step2.root.unused (for future reference)

image

PR #384, opened 2021-11-22
Enable backward search for pixel-quadruplet-seeded iterations

mmasciov:
As of today, only for testing.
Please, DO NOT MERGE (yet).

PR #387, opened 2021-11-24
Enable backward search for iterations seeded by pixel quadruplets

mmasciov:
### PR description:

This PR is a further response following BTV investigations and aims at recovering as many pixel hits as possible compared to CKF, in the attempt to further improve impact parameter resolution. In summary:
(1) it enables backward search for iterations seeded by pixel quadruplets;
(2) it tightens the seed duplicate cleaning/merging.

For (1), first two BPIX layers in mkFit layer plan are inverted, only for iterations seeded by pixel quadruplets and only in transition regions and endcaps. In addition, hit selection windows are set also in pixel layers, to allow for backward search.
(2) allows for a further reduction in fakes+duplicates, with no cost in efficiency and with an increase of merged pixel hits in endcaps.

It supersedes PR #384.

### PR validation (MTV results):

- TTbar, PU=50 (vs. CKF and vs. PR386): https://mmasciovecchio.web.cern.ch/BTV_Nov2021/MTV_PR386_plusBkwSearchPixelQuad_TTbarPU50_122X/

- 10muPt0p2to10 (vs. CKF and vs. PR386): https://mmasciovecchio.web.cern.ch/BTV_Nov2021/MTV_PR386_plusBkwSearchPixelQuad_pixelLessSeedCleaning_10muPt0p2to10_122X/

- High-pT QCD (cross-check vs. CKF and vs. mkFit as in 12_2_0_pre2): https://mmasciovecchio.web.cern.ch/BTV_Nov2021/MTV_highPtQCD_mkFit_for1220pre3/


mkFit/MkFinder.cc
```diff
@@ -1174,13 +1192,30 @@ void MkFinder::FindCandidatesCloneEngine(const LayerOfHits &layer_of_hits, CandC
...
+		if(i>2) break;
+		if(ccand.m_hots[i].m_hot.layer==layer_of_hits.layer_id())
+		{
+		  hitExists = true;
+		  break;
+		}
+	      }
+	    }
+	    if (hitExists)
+	      continue;
```
slava77, line 1211: in view of https://github.com/trackreco/cmssw/pull/162
I wanted to understand a bit better what this code does.
Why does the check stop at 2 (checking only the first 3 hits): was it supposed to be the same as the number of seed hits?
Is it really a problem to have more than one hit per layer (as long as it's not the same module)?
Also, why only pixel layers are checked?

slava77: @mmasciov

mmasciov: I don't recall the details, maybe @osschar does?
From the description of the PR, this was in the context of the BTV validation. I would therefore assume it had the purpose to forbid the addition of more hits on the same layer only for the first pixel layers, which would motivate the truncation at 2 in the loop over the number of hits.

osschar:
I was staring at this for a bit. I think this was put in as a band-aid after flipping the layer order for first and second layer to recover the missing hit on the second layer -- and therefore it just checks if the first layer is already there from the seed, then don't try to add anything else. i < 2 is probably heuristic, from the highest number of hits after the seed merging -- or maybe it also prevents the addition of a new hit on the second layer, in case it was actually already there.

PR #393, opened 2022-02-02
modify find_duplicates_sharedhits 

leonardogiannini:
this PR modifies find_duplicates_sharedhits used in pixelless and tobtec to fdollow a logic similar to cmssw
1) maps hits to tracks
2) find for each tracks the ones the share hits with it and how many

this allows to avoid nested track-track loops

the timing is reduced by a factor from ~2 to 4, while the physics is not changed  i.e. same number of output tracks

- new vs old time for the function (10 evts)
image

- speed up plot (10 evts)
image


mmasciov:
Closing, as superseded by cms-sw/cmssw#37122.

